PR #39, opened 2014-11-21
read_big_query() added

Mullefa:
I have found this function to be very convenient when working with either large queries, or lots of queries, or queries that can be parametrised by some variable(s).


krlmlr:
I'm not sure this belongs in bigrquery.

Mullefa:
Agreed. I was going to suggest the [SQL](https://github.com/rstats-db/SQL) package, but saw it has been deprecated. Doesn't seem like DBI is the place to put it either though?!

krlmlr:
In the future, DBI might offer support for parametrized queries (rstats-db/DBI#63), even if the backend doesn't support it. I remember that similar functionality is available in the ProjectTemplate package, too.

hadley:
Yeah, this doesn't belong in bigrquery, sorry.

PR #44, opened 2015-01-05
Cannot use max_results in function list_tables()

craigcitro:
When there are many tables in the dataset, I cannot retrieve a full list of them without option max_results. But after I added this parameter, got this error:

```
> histTxnTables <- list_tables(bqdaproject, bqBaseDataset)
Auto-refreshing stale OAuth token.
> histTxnTables <- list_tables(bqdaproject, bqBaseDataset, 1000)
Error: HTTP error [404] Not Found
```


craigcitro:
PTAL @hadley

PR #46, opened 2015-01-20
Add a new function to copy table

craigcitro:
It will be very useful to have a copy table function in some cases.
https://cloud.google.com/bigquery/docs/tables#copyingtable

Thanks in advance!


craigcitro:
PTAL @hadley 

the code here was easy to write -- the hard part is (as always) naming. ;) some questions:

* we've got a source table with three-part identifier, and the same for the destination. here i've assumed that the common case is that the project and dataset match, though maybe i'm crazy?
* i didn't expose `createDisposition` and `writeDisposition` -- should i add those?
* interestingly, a copy can take one or *many* source tables. should i "clean up" the API by taking a vector or list of source tables, and simplify the argument names on the destination table?

totogo:
In our use cases, at least  ```writeDisposition``` is a very useful parameter. I strongly hope you could expose these options. Actually its funny that I need the copy function because ```writeDisposition``` is not exposed in query job. 
In my case, we want to add some new columns in table A. So in current ```bigrquery``` I need run a query, join table A and B, then use the result to replace A. But since table A already exists, I got an error. I have to put the join result in table C, delete table A, then list table C to A.
As you can see if ```writeDisposition``` and ```createDisposition``` can be exposed in the jobs, it can save us some cost in BigQuery.

I wrote this function in our script FYI:

```R

#' Create a new job to copy an existing table
#'
#' This is a low-level function that creates a copy job. To wait until it is
#' finished and then retrieve the results, see \code{\link{copy_table}}
#'
#' @param source_table source table to copy,
#'   either as a string in the format used by BigQuery, or as a list with
#'   \code{project_id}, \code{dataset_id}, and \code{table_id} entries
#' @param destination_table destination table for copied table,
#'   either as a string in the format used by BigQuery, or as a list with
#'   \code{project_id}, \code{dataset_id}, and \code{table_id} entries
#' @param project project id
#' @param create_disposition (Optional) Specifies the action that occurs
#'   if the destination table already exists.
#'   The following values are supported:
#'   \itemize{
#'     \item \code{CREATE_IF_NEEDED}: If the table does not exist, BigQuery
#'     creates the table.
#'     \item \code{CREATE_NEVER}: The table must already exist. If it does not,
#'     a 'notFound' error is returned in the job result.
#'   }
#'   The default value is \code{CREATE_IF_NEEDED}.
#' @param write_disposition (Optional) Specifies the action that occurs
#'   if the destination table already exists.
#'   The following values are supported:
#'   \itemize{
#'     \item \code{WRITE_TRUNCATE}: If the table already exists,
#'     BigQuery overwrites the table data.
#'     \item \code{WRITE_APPEND}: If the table already exists, BigQuery appends
#'     the data to the table.
#'     \item \code{WRITE_EMPTY}: If the table already exists and contains data,
#'     a 'duplicate' error is returned in the job result.
#'   }
#'   The default value is \code{WRITE_EMPTY}.
#' @family jobs
#' @return a job resource list, as documented at
#'   \url{https://developers.google.com/bigquery/docs/reference/v2/jobs}
#' @seealso API documentation for insert method:
#'   \url{https://developers.google.com/bigquery/docs/reference/v2/jobs/insert}
#' @export
insert_copy_job <- function(source_table, destination_table, project,
                            write_disposition = "WRITE_EMPTY") {
  if (is.character(source_table)) {
    source_table <- parse_table(source_table, project_id = project)
  }
  assert_that(is.string(source_table$project_id),
              is.string(source_table$dataset_id),
              is.string(source_table$table_id))

  if (is.character(destination_table)) {
    destination_table <- parse_table(destination_table, project_id = project)
  }
  assert_that(is.string(destination_table$project_id),
              is.string(destination_table$dataset_id),
              is.string(destination_table$table_id))


  url <- sprintf("projects/%s/jobs", project)
  body <- list(
    configuration = list(
      copy = list(
        sourceTable = list(
          projectId = source_table$project_id,
          datasetId = source_table$dataset_id,
          tableId = source_table$table_id),
        destinationTable = list(
          projectId = destination_table$project_id,
          datasetId = destination_table$dataset_id,
          tableId = destination_table$table_id)
      )
    )
  )

  if (create_disposition == "CREATE_IF_NEEDED") {
    body$configuration$copy$createDisposition <- "CREATE_IF_NEEDED"
  } else if (create_disposition == "CREATE_NEVER") {
    body$configuration$copy$createDisposition <- "CREATE_NEVER"
  } else {
    #error
  }

  if (write_disposition == "WRITE_EMPTY") {
    body$configuration$copy$writeDisposition <- "WRITE_EMPTY"
  } else if (write_disposition == "WRITE_APPEND") {
    body$configuration$copy$writeDisposition <- "WRITE_APPEND"
  } else if (write_disposition == "WRITE_TRUNCATE") {
    body$configuration$copy$writeDisposition <- "WRITE_TRUNCATE"
  } else {
    #error
  }

  bq_post(url, body)
}
```

craigcitro:
@totogo in that case, would it be easier if i added this, but also exposed those options in the query calls?

craigcitro:
(i'm going to go another round with my code, i'll ping this again before it's ready for review)

craigcitro:
finally got around to updating this one. PTAL @hadley

R/tables.r
```diff
@@ -72,3 +72,76 @@ delete_table <- function(project, dataset, table) {
...
+#' copy_table(list(src, dest), doubled)
+#' }
+copy_table <- function(src, dest,
+                       create_disposition = "CREATE_IF_NEEDED",
+                       write_disposition = "WRITE_EMPTY",
+                       project = NULL) {
+  if (validate_table_reference(src)) {
+    src <- list(src)
+  } else if (!all(vapply(src, validate_table_reference, TRUE))) {
+    stop("src must be a table reference or a list of table references")
```
hadley: Maybe this should go in `validate_table_references()`? I normally call functions of that type `check_*`

hadley: Oh I see you're trying to give more informative messages :)

R/tables.r
```diff
@@ -72,3 +72,76 @@ delete_table <- function(project, dataset, table) {
...
+  }
+  if (!validate_table_reference(dest)) {
+    stop("dest must be a table reference")
+  }
+  project = project %||% dest$project_id
```
hadley: `=` -> `<-`

hadley:
Few minor comments. Do you know why the build is failing?

craigcitro:
just noticed the build failure ... which was a bug. (it's interesting to me that i didn't see it before i pushed, it pops up all the same locally.)

R/tables.r
craigcitro: done

R/tables.r
craigcitro: yep. at once point i thought about also including info on how the table reference was malformed, but that got a little heavy.

ericbrownaustin:
Was the writeDisposition parameter added to the function?  It would be very useful as it would allow users to create a table directly in r without having to manually create an empty table in the BQ console.

craigcitro:
@ericbrownaustin Can you say more about what you're looking to do? There are disposition arguments on copy, but in general you can just create the table as part of the load/copy/query operation.

PR #49, opened 2015-04-29
Add support for reading API keys from an env var.

craigcitro:
This adds support for API keys to bigrquery, reading them from the
`BIGRQUERY_API_KEY` environment variable. They're used in all API calls when
present.

PTAL @hadley

### DIFF
--- a/R/request.r
+++ b/R/request.r
@@ -1,30 +1,44 @@
 base_url <- "https://www.googleapis.com/bigquery/v2/"
 upload_url <- "https://www.googleapis.com/upload/bigquery/v2/"
 
+prepare_bq_query <- function(query) {
+  api_key <- Sys.getenv("BIGRQUERY_API_KEY")
+  if (!nzchar(api_key)) {
+    return(query)
+  }
+  query <- query %||% list()
+  query[["key"]] <- query[["key"]] %||% api_key
+  query
+}
+
 #' @importFrom httr GET config
-bq_get <- function(url, ..., token = get_access_cred()) {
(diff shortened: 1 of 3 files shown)


R/request.r
```diff
@@ -1,30 +1,47 @@
 base_url <- "https://www.googleapis.com/bigquery/v2/"
 upload_url <- "https://www.googleapis.com/upload/bigquery/v2/"
 
+prepare_bq_query <- function(query) {
+  api_key <- Sys.getenv("BIGRQUERY_API_KEY")
+  if (nzchar(api_key)) {
```
hadley: Tiny niggles: I think this would be easier to understand if you did `if (!nzchar(api_key)) return(query)` etc

craigcitro: done -- in fact, shortened a good bit when i remembered `%||%`.

feel free to spot more niggles. :smirk:

craigcitro:
PTAL

hadley:
YEEEEESSSS.

(Can you please add a bullet point to news before merging?)

AspireToCodeBetter:
@craigcitro is this been merged? for some reason i am unable to use big query and it still asks for browser authentication.

I am setting the env variable as below.

```r
Sys.setenv(BIGRQUERY_API_KEY="TuxYqUc3eVzkllXnMHbk2iIZSp6wZASkj9nNsDkyX")
```

byapparov:
Where `BIGRQUERY_API_KEY` can be found in the Google API Console? Is it the only environment variable required for access to bigquery?

PR #51, opened 2015-06-17
Added db_save_query.bigquery so compute() now works

akhmed1:
Hello everyone! This pull request features an implementation of `db_save_query` for "bigquery". This is my first attempt to contribute to `bigrquery` project so I apologize in advance if I am missing some key details (or may be even being overly verbose in describing it)

### Motivation
The motivation for this pull request is that the current implementation of `bigrquery` allows one to use `collect` function (that is, saving the results locally) but not `compute` function (that is, saving the results remotely).

`Compute` function seems to be extremely important when working with big datasets as it allows one to keep big temporary remotely results in Google BigQuery without downloading them locally and use them in subsequent queries.

### Implementation Details

* `compute` function fails only because of the lack of `db_save_query.bigquery`.
Fortunately, `query_exec` facilities allow saving the query result and thus, `db_save_query` could potentially be implemented as a hook to query_exec with a particular `destination_table`.

* `temporary = TRUE` is not currently implemented and is currently ignored so all tables are created as persistent.

* This pull request works correctly with the suggested `dplyr` pull request [hadley/dplyr#1220](https://github.com/hadley/dplyr/pull/1220) which also implements JOINs inside BigQuery. Together these two pull requests can extend the current functionality to allow both JOINing remotely and storing the results remotely.

### Tests

The code belows demonstrates a simple test of `compute` function. 
``` r
library(dplyr)
library(bigrquery)

BILLINT_DATASET <- "<DATASET_INSIDE_THAT_PROJECT>"

test_df <- copy_to(bq, df, name = "test_df")

# Perform a simple query and save the result remotely
test_df2 <- test_df %>%
              summarise( n = n() ) %>%
              compute( name = "test_df2", temporary = FALSE)

# Display the result
test_df2 %>% collect
```

Thank you for consideration!


akhmed1:
Moving this pull request to a separate branch. Please see new pull request.

akhmed1:
Moving this pull request to a separate branch. Please see new pull request.

PR #55, opened 2015-07-03
mutate_ always requires collapsing in BigQuery

akhmed1:
This pull request fixes the issue with `mutate` in BigQuery when the variables created with `mutate` cannot be used in subsequent operators (as explained below).

### Motivation
BigQuery does not allow one to use column alias in the same query where it is assigned as explained [in the reference](https://cloud.google.com/bigquery/query-reference#select), and therefore, the variable created by `mutate_` cannot be used until one collapses the SQL query first.

Knowing this, it may be beneficial in Google BigQuery to collapse the query immediately after mutate so that the very next operator can use the newly created variable.

### Tests

The code belows demonstrates the failure of regular `mutate_`
``` r
library(dplyr)
library(bigrquery)

mutate_df <- copy_to(bq, df, name = "mutate_df")

# Perform two consecutive mutate queries 
# where 2nd mutate uses the variable created in the 1st mutate
mutate_df %>%
  mutate( c = a + 1 ) %>%
  mutate( d = c + 1 ) %>% 
  explain
```
This produces the following query
``` sql
SELECT [a], [b], [a] + 1.0 AS [c], [c] + 1.0 AS [d]
FROM [mutate_df]
```
which is not allowed in BigQuery as per [reference](https://cloud.google.com/bigquery/query-reference#select) and will return an error:
```
Error: Field 'c' not found in table 'mutate_df'; did you mean 'a'?

query invalidQuery. Field 'c' not found in table 'mutate_df'; did you mean 'a'?
```

### Contribution
This pull request makes sure to collapse the query once mutate is used. This allows one to use the newly created names in the subsequent operators. With pull request, the same code as before:
``` r
mutate_df %>%
  mutate( c = a + 1 ) %>%
  mutate( d = c + 1 ) %>% 
  explain
```
produces
``` sql
SELECT [a], [b], [c], [d]
FROM (SELECT [a], [b], [c], [c] + 1.0 AS [d]
FROM (SELECT [a], [b], [a] + 1.0 AS [c]
FROM [mutate_df]) AS [_W1]) AS [_W2]
```
which is now a legitimate BigQuery query that produces the correct answer.

Thank you for consideration!


R/dplyr.r
```diff
@@ -99,6 +99,20 @@ dim.tbl_bigquery <- function(x) {
   c(NA, p)
 }
 
+#' @export
+#' @importFrom dplyr mutate_
+mutate_.tbl_bigquery <- function(.data, ..., .dots) {
```
craigcitro: why not just call `dplyr:::mutate_.tbl_sql`, and then call `collapse`? (is there a reason we need to reproduce it here?

akhmed1:
@craigcitro Thanks for the feedback! My concern with just collapsing a result of `dplyr:::mutate_.tbl_sql` was that some queries (say, with window functions) do get collapsed inside `mutate_.tbl_sql` while others don't. So the intent was to avoid double-collapse and extra-level of nesting for those queries (at the expense of code re-use). I definitely see the benefits of reuse approach too and I can update the pull request if that's the desired solution. Thanks!

craigcitro:
the bigquery query planner is smart enough to drop the extra nesting, so I think in this case it's a little cleaner to reuse the existing code ...

akhmed1:
Then the re-use solution looks definitely better. Just updated the pull request!

craigcitro:
looks great -- thanks!

PR #61, opened 2015-11-26
Sketch of a credential fetch registration system.

craigcitro:
This is still quite rough, but it's a stab at trying to support credential
registration in bigrquery (which we'd potentially want to move into httr?).
This would allow us to have packages provide their own credential fetch
mechanisms which get used in the default "get an access token" path; two
important use cases for this:

* For Travis, we could have the travis package simply register a new
  credential mechanism that fetches from the environment. Now travis
  credentials can be used in many packages just by adding a dependency and
  loading it (assuming we add some code in `zzz.R` for travis).

* For systems like AWS EC2 and GCE, there's a notion of credentials associated
  with the machine itself (GCE calls this a service account, I forget if AWS
  uses the same term). We could add a hook (here or in another package)
  to **transparently** use those credentials when available.

Note that this is just a sketch, I'm looking to incite some conversation. There are still questions to be handled (how does one control order being a big one).

Once we get closer, feel free to pick apart how much my R looks like I've been
writing Python all day. :wink: In particular, I clearly modeled this as something I'd use with a decorator in Python; what's the idiomatic way to do that in R?

Adding several folks for conversation: @deflaux @jennybc @hadley @jeroenooms

### DIFF
--- a/R/auth.r
+++ b/R/auth.r
@@ -27,15 +27,20 @@ bq_env <- new.env(parent = emptyenv())
 get_access_cred <- function() {
   cred <- bq_env$access_cred
   if (is.null(cred)) {
-    cred <- oauth2.0_token(google, bigqr,
-      scope = c(
-          "https://www.googleapis.com/auth/bigquery",
-          "https://www.googleapis.com/auth/cloud-platform"))
-
-    # Stop if unsuccessful
-    set_access_cred(cred)
+    scopes <- c(
+        "https://www.googleapis.com/auth/bigquery",
+        "https://www.googleapis.com/auth/cloud-platform")
+    for (f in bq_env$credential_fetchers) {
+      cred <- f(google, bigqr, scopes)
+      if (!is.null(cred)) {
+        set_access_cred(cred)
+        break
+      }
+    }
   }
-
+  if(is.null(cred)) {
+    stop("Failed to create OAuth2.0 credentials for executing BigQuery request.")
+  }
   cred
 }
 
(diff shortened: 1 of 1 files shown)


jennybc:
cc @richfitz @karthik @sckott @cboettig as this is related to a topic rOpenSci is thinking about

karthik:
Thanks @jennybc 
Following the conversation now.

sckott:
thanks jenny

nathangoulding:
To add to the rough sketch of this, something like this for the service token might be nice:

```
fetch_service_token_creds <- function() {
    if (!is.string(bq_env$service_token_file) || length(bq_env$service_token_file) == 0) {
        stop("call set_service_token with a valid filename")
    }

    endpoint <- httr::oauth_endpoints("google")

    secrets <- jsonlite::fromJSON(bq_env$service_token_file)

    scope <- c(
        "https://www.googleapis.com/auth/bigquery",
        "https://www.googleapis.com/auth/cloud-platform")

    httr::oauth_service_token(endpoint, secrets, scope)
}

set_service_token <- function(token_file) {
    bq_env$service_token_file = token_file
    register_credential_fetcher(fetch_service_token_creds)
}
```

Usage would then simply be:

```
bigrquery:: set_service_token("/path/to/token.json")
```

nathangoulding:
Implemented PoC of this here: https://github.com/craigcitro/bigrquery/pull/1

One thing I'm wondering about is the merits of registering multiple auth callbacks for the automatic determination of which creds to use. Would this be potentially better suited to a simpler `set_credential_fetcher` that defaults to the standard interactive 3LO instead of multiple callbacks?

This way the service token or travis CI mechanism would never get "stuck" behind the interactive browser-based flow that could potentially stop execution because of an error like `interactive() is not TRUE`, e.g. https://github.com/hadley/bigrquery/issues/42.

nathangoulding:
You do call out order being the most important TBD factor here, but I might argue that support for multiple callbacks is typically better suited for bootstrapping, or a provider informing an implementer that a thing just happened, e.g. onAccessCredsFetched - where the implementer can't really disrupt the execution flow in their callback.

Here, we just need to know how to perform our access credentials fetching, which might be better achieved through to a single callback. 

Thoughts?

craigcitro:
For anyone following along: after a bunch of offlist discussion, we figured out that the right home for this would actually be in its own repo, which could then be used by bigrquery, google genomics, and others. There's now a [repo](https://github.com/ropenscilabs/gauth), which should start seeing updates shortly (once I'm back at work).

nathangoulding:
Sounds good. :+1: Let me know if you want any help with `gauth`.

craigcitro:
OK, still not there yet, but https://github.com/ropenscilabs/gauth is at least starting to get a little code.

Closing this -- all the action's happening there.

PR #63, opened 2015-12-21
rely on http status code instead of parsed text

nathangoulding:
This improves robustness by checking the HTTP status code instead of relying on the status text, related to https://github.com/hadley/httr/pull/305.


R/request.r
```diff
@@ -48,7 +48,7 @@ process_request <- function(req) {
   # No content -> success
   if (req$status_code == 204) return(TRUE)
 
-  if (http_status(req)$category == "success") {
+  if (req$status_code %/% 100 == 2) {
```
craigcitro: two thoughts:

1. I think we want to call anything lower than a 400 a success?
2. It's probably more readable to just check if the status is >= 200 and < 400 (or 300, cf previous comment).

nathangoulding: 1. Beyond 304 not modifed, treating 3xx as a success here might produce inconsistent results, and my thinking was to mimic the behavior of httr's response-status.r for maximum compatibility, which only treats 2xx as success.
2. httr's response-status.r does the `/ 100` magic, but I definitely prefer to err on the side of readability and will make the worthy update.

R/request.r
```diff
@@ -48,7 +48,7 @@ process_request <- function(req) {
   # No content -> success
   if (req$status_code == 204) return(TRUE)
 
-  if (http_status(req)$category == "success") {
+  if (req$status_code >= 200 && req$status_code < 300) {
```
hadley: I think it would be better to use `httr::status_code(req)` or take a dependency on the dev version and use `!httr::http_error(req)`

nathangoulding: Do we currently get something from `httr::status_code` that we can safely rely on? The `category` recently got capitalized which is why I wanted to switch this to check the more immutable HTTP code.

craigcitro: `httr::status_code` will always return an int, so i don't think there's any worry that it'll have something comparable to case sensitivity issues, right?

as far as consistency on the API side, changing the response code amounts to an API change, so it *shouldn't* happen.

and, on the note of what values we care about, the [bigquery error docs](https://cloud.google.com/bigquery/troubleshooting-errors) definitely suggest `>= 400` is an error. this matches `httr::http_error`, which is no coincidence: the [rfc for status codes](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) says 4xx is a client error, 5xx is a server error; for google APIs, in general this means 4xx shouldn't be retried, and 5xx should (with backoff).

nathangoulding: You're right about `status_code` returning an int. I was looking at `http_status` still, sorry. I'll make that update shortly.

`httr` is correct in implementing `>= 400` as a sign of HTTP error of course. Here I think we may be more interested in whether from bigrquery's perspective we should treat any non-2xx code as a sign of an error. There's a fine argument to make that it would be odd to receive a 301 or 302 from bigquery and treat it as a success. Currently `bigrquery` works this way, as `http_status(req)$category` only returns `success` for `2xx`.

Should we continue to treat non-2xx as an error in bigrquery as we move to use `status_code`, or should we change this behavior and start treating 1xx and 3xx as success? I would lean towards keeping behavior consistent with how it currently works (only 2xx we treat as success), but I'll defer to you guys of course.

PR #64, opened 2015-12-22
set encoding to utf8

nathangoulding:
httr logs to stderr if the encoding isn't set when calling `guess_encoding` from the `parse_` methods. This cleans up this error by explicitly setting it to UTF-8, which is what httr defaults it to if it's not specified.

### DIFF
--- a/R/request.r
+++ b/R/request.r
@@ -49,12 +49,12 @@ process_request <- function(req) {
   if (req$status_code == 204) return(TRUE)
 
   if (http_status(req)$category == "success") {
-    return(content(req, "parsed", "application/json"))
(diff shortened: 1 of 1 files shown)


nathangoulding:
It looks like this started logging to stderr last Thursday 12/17 when the default encoding in httr was changed from `ISO-8859-1` to `UTF-8`.

nathangoulding:
If this change is something where we should maintain the `ISO-8859-1` compatibility with what it was previously, we could default it to `ISO-8859-1` and expose the character set via `bq_env` potentially. But that might be overkill.

hadley:
This should be fixed in httr, because `application/json` is always UTF-8

PR #65, opened 2015-12-22
use status_code for success determination

nathangoulding:
This improves robustness by checking the HTTP status code instead of relying on the status text, related to hadley/httr#305.


hadley:
Thanks!

PR #66, opened 2015-12-22
implement crud for datasets

nathangoulding:
This implements the basic CRUD operations for datasets beyond `list_datasets`, so you don't have to jump back out to the bigquery console to create/delete datasets. I didn't go nuts with all of the properties of the Dataset resource, opting instead to keep it simple for what I think is probably the most common use case.


R/datasets.r
```diff
@@ -18,3 +19,106 @@ list_datasets <- function(project) {
 
   unlist(lapply(data, function(x) x$datasetReference$datasetId))
 }
+
+#' Gets an existing dataset in a project
+#'
+#' @param project The project name, a string
```
craigcitro: let's do `project` for consistency with `datasetId`?

craigcitro: ... and looking back, i think we decided to go with just `project`/`dataset`/`table` in `tables.r`. so maybe let's go the other way -- `project` and `dataset`?

(the only hesitation i have is that the API itself uses the `Id` suffix everywhere.)

R/datasets.r
```diff
@@ -18,3 +19,106 @@ list_datasets <- function(project) {
...
+  url <- sprintf("projects/%s/datasets/%s", project, datasetId)
+  bq_delete(url, query = list(deleteContents = deleteContents))
+}
+
+#' Creates a new dataset in a project
+#'
+#' @param project The project name, a string
+#' @param datasetId The datasetId to delete, a string
+#' @param description The dataset description, a string
+#' @param friendlyName The dataset's friendly name, a string
```
craigcitro: just to ask -- do people really use `friendlyName`? (i thought it wasn't even exposed in the UI.)

R/datasets.r
```diff
@@ -18,3 +19,106 @@ list_datasets <- function(project) {
...
+#' @seealso Google API documentation:
+#'   \url{https://cloud.google.com/bigquery/docs/reference/v2/datasets/delete}
```
craigcitro: update

R/datasets.r
```diff
@@ -18,3 +19,106 @@ list_datasets <- function(project) {
...
+#' @param project The project name, a string
+#' @param datasetId The datasetId to delete, a string
+#' @param deleteContents Whether to delete the tables if the dataset is not empty, a boolean
+#' @seealso Google API documentation:
+#'   \url{https://cloud.google.com/bigquery/docs/reference/v2/datasets/delete}
+#' @family datasets
+#' @export
+#' @examples
+#' \dontrun{
+#' delete_dataset("publicdata", "shakespeare", TRUE)
```
craigcitro: for examples, i think it's worth using the name explicitly for optional arguments

craigcitro:
a few little nits, but this looks good.

one bigger question: is there anything useful we could do as a test here? (having at least *some* test that imports this code is useful, even if it's a mostly trivial test.)

R/datasets.r
nathangoulding: Ahh yeah, we should probably stick with `project` and `dataset` in favor of consistency with what's been done before in `tables.r` even though it doesn't match the API.

R/datasets.r
nathangoulding: Roger that.

R/datasets.r
nathangoulding: I didn't know it existed until I went to implement this TBH.

Interestingly enough (or not), when you rename a project in the Google Developer Console, you're actually editing the `friendlyName` field of the project. The bigquery UI exposes the `datasetId` everywhere, and so that's the only thing you're ever really aware of. So at some point we may see an update to the bigquery UI that adds the ability to "rename a dataset", and this would presumably be allowing you to edit the `friendlyName`, while continuing to keep the `datasetId` safely immutable.

R/datasets.r
nathangoulding: Good catch.

nathangoulding:
OK I think I've addressed all your comments. I ended up creating validation tests for datasets, which will at least ensure that there is some basic coverage.

craigcitro:
LGTM -- @hadley do you have strong feelings about squashing commits?

nathangoulding:
Closing this in favor of a rebased https://github.com/hadley/bigrquery/pull/69.

PR #67, opened 2015-12-23
upgrade roxygen2

krlmlr:


craigcitro:
does this need to get re-run when we merge @nathangoulding's PR?

krlmlr:
Re-documenting is a good idea. There probably will be only whitespace changes (if any) in the autogenerated files, though.

PR #70, opened 2015-12-30
service account support

nathangoulding:
This adds support for the service account for use with BigQuery. This performs 2LO when you call `set_service_token`, as an alternative to the credentials registration mechanism as outlined here: https://github.com/rstats-db/bigrquery/pull/61.

Usage is simple:

```
bigrquery::set_service_token("/path/to/token.json")
bigrquery::list_projects()
```

For comparison, I implemented service account support using the previously outlined credentials registration mechanism as well, here: https://github.com/craigcitro/bigrquery/pull/1

I personally prefer this simpler and more declarative approach, but I'll defer of course.


hadley:
This looks good to me.  Once this is merged I can add an encrypted service token to the repo so we can do unit testing on travis (cc @krlmlr)

@craigcitro what do you think?

craigcitro:
Short term, if this gets people unblocked, let's add it. My one caveat: I'd prefer not to `@export` this function, because I'd rather people depend on a shared auth library rather than this particular method in `bigrquery`.

Longer term, I think we will want to move to something akin to what I described in #61, but it'll live in its own package. Once I'm back from vacation, I'll get that repo fleshed out, and we can see what it'd look like for use here?

nathangoulding:
Not exporting `set_service_token` sounds fine to me if the long-term vision is that all the google auth will be managed through a separate package. We can still use `bigrquery:::set_service_token("token.json")` for travis in the meantime, or for anyone that needs this functionality now. Then we can safely swap this out when the future state is ready.

@hadley Sound good to you? If so, I'll do the needful.

hadley:
Yeah, that seems reasonable to me.

nathangoulding:
Closing in favor of an exported and rebased version.

PR #71, opened 2015-12-30
internal service token support

nathangoulding:
This adds internal support for the service account for use with BigQuery.

```
bigrquery:::set_service_token("/path/to/token.json")
bigrquery::list_projects()
```

### DIFF
--- a/R/auth.r
+++ b/R/auth.r
@@ -55,18 +55,10 @@ get_sig <- function() {
   stop("Deprecated: use get_access_cred directly", call. = FALSE)
 }
 
-set_service_token <- function(token_file) {
+# The service_token can be an absolute or relative filename, or JSON text
+set_service_token <- function(service_token) {
 
-  # If the file doesn't exist, jsonlite::fromJSON will attempt to parse
-  # the string literal, which throws a less-than-intuitive error to the
-  # user of bigrquery::set_service_token
-  if (!file.exists(token_file)) {
-    stop("Invalid service token file; file does not exist")
-  }
-
-  # This will throw an error if we don't have permissions to read the
-  # file or if the file contains invalid JSON.
-  service_token <- jsonlite::fromJSON(token_file)
+  service_token <- jsonlite::fromJSON(service_token)
 
   endpoint <- httr::oauth_endpoints("google")
 
--- a/R/auth.r
+++ b/R/auth.r
@@ -54,3 +54,26 @@ reset_access_cred <- function() {
 get_sig <- function() {
   stop("Deprecated: use get_access_cred directly", call. = FALSE)
 }
+
+set_service_token <- function(token_file) {
+
+  # If the file doesn't exist, jsonlite::fromJSON will attempt to parse
+  # the string literal, which throws a less-than-intuitive error to the
+  # user of bigrquery::set_service_token
+  if (!file.exists(token_file)) {
+    stop("Invalid service token file; file does not exist")
+  }
+
+  # This will throw an error if we don't have permissions to read the
+  # file or if the file contains invalid JSON.
+  service_token <- jsonlite::fromJSON(token_file)
+
+  endpoint <- httr::oauth_endpoints("google")
+
+  scope <- "https://www.googleapis.com/auth/bigquery"
+
+  cred <- httr::oauth_service_token(endpoint, service_token, scope)
+
+  set_access_cred(cred)
+}
+


R/auth.r
```diff
@@ -54,3 +54,26 @@ reset_access_cred <- function() {
 get_sig <- function() {
   stop("Deprecated: use get_access_cred directly", call. = FALSE)
 }
+
+set_service_token <- function(token_file) {
+
+  # If the file doesn't exist, jsonlite::fromJSON will attempt to parse
+  # the string literal, which throws a less-than-intuitive error to the
+  # user of bigrquery::set_service_token
+  if (!file.exists(token_file)) {
```
krlmlr: Why don't you simply allow strings here:

```r
bigrquery:::set_service_token(Sys.getenv("GOOGLE_TOKEN"))
```

```bash
Rscript -e "bigrquery:::set_service_token('${GOOGLE_TOKEN}')"
```

Removes the need for a file stored in the repo.

nathangoulding: This was really just so that an end user wouldn't be left with a JSON parsing error if they provided an invalid filename. 

Since this function has been unexported and it sounds like users will be using the future shared google auth library to do this, I'm happy to remove this check.

@hadley @craigcitro You guys agree?

nathangoulding:
I went ahead and made the change to remove the `file.exists` check so it now supports passing in a JSON string directly.

nathangoulding:
@hadley @craigcitro @krlmlr Is this good to go from your perspectives?

krlmlr:
How do I use it?

I revoked bigrquery permissions from my Google account (https://github.com/hadley/httr/issues/311), then called `set_service_token("~/Downloads/bigrquery-2ad41c8d1c43.json")` (passing the file that was downloaded when I created a new service account key via https://console.developers.google.com/apis/credentials). This succeeded, but I still can't run queries:

```
> query_exec(sql, project = project)
Error: is.request(y) is not TRUE
16: stop(sprintf(ngettext(length(r), "%s is not TRUE", "%s are not all TRUE"), 
        ch), call. = FALSE, domain = NA)
15: stopifnot(is.request(x), is.request(y))
14: f(init, x[[i]])
13: Reduce(request_combine, list(...))
12: c.request(req, token$sign(req$method, req$url))
11: c(req, token$sign(req$method, req$url))
10: request_prepare(req)
9: request_perform(req, hu$handle$handle)
8: POST(paste0(base_url, url), body = json, config(token = token), 
       add_headers(`Content-Type` = "application/json"), ..., query = prepare_bq_query(query)) at request.r#31
7: bq_post(url, body) at jobs.r#60
6: insert_query_job(query, project, destination_table = destination_table, 
       default_dataset = default_dataset) at query.r#32
5: query_exec(sql, project = project) at .active-rstudio-document#4
4: eval(expr, envir, enclos)
3: eval(ei, envir)
2: withVisible(eval(ei, envir))
1: source("~/.active-rstudio-document", echo = TRUE)
```

nathangoulding:
@krlmlr That error is related to an unresolved issue in `httr`, and some discussion around it can be found here:  https://github.com/hadley/httr/issues/275.

I haven't yet had a chance to find out why the `request` is getting passed in to `request_combine` as `y$config` instead of `y`, but this one-liner to `httr/R/request.R` will allow you to test this PR: https://github.com/hadley/httr/pull/304/files.

nathangoulding:
@krlmlr FWIW, I did a little more digging and this is the better fix: https://github.com/hadley/httr/pull/313

krlmlr:
Works for me with https://github.com/hadley/httr/pull/313, although I'm now seeing new warnings (which may well be due to the httr update):

```
> query_exec(sql, project = project)
No encoding supplied: defaulting to UTF-8.
No encoding supplied: defaulting to UTF-8.
No encoding supplied: defaulting to UTF-8.
No encoding supplied: defaulting to UTF-8.
  year month day weight_pounds
1 1969     2  14      7.125340
2 1969     8  17      6.876218
3 1969     5  31      7.561856
4 1969     4  18      8.000575
5 1969     1   4      8.437091
```

I still observe https://github.com/hadley/httr/issues/311.

nathangoulding:
Correct, the UTF-8 warnings are the result of a change in `httr` that sets the default charset to `UTF-8` in the dev version.

I haven't looked at https://github.com/hadley/httr/issues/311 yet but I assume there's no logic to remove the local cache in the case of revoked credentials (just a guess and needs some digging / further work).

nathangoulding:
@hadley @krlmlr Anything else here before we merge this in?

craigcitro:
sorry for the delay! :grinning:

PR #72, opened 2016-01-06
rename to rstats-db

nathangoulding:
Following the move, updates occurrences of `hadley/bigrquery` to `rstats-db/bigrquery`.


hadley:
Thanks!

PR #76, opened 2016-01-13
set default options in .onLoad() instead of .onAttach()

krlmlr:
to support is_quiet() when loading package without attaching

Should the option be renamed to `bigrquery.quiet` (now: `bigquery.quiet`)?

### DIFF
--- a/R/zzz.r
+++ b/R/zzz.r
@@ -1,4 +1,4 @@
-.onAttach <- function(libname, pkgname) {
+.onLoad <- function(libname, pkgname) {
   op <- options()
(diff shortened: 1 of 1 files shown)


krlmlr:
@craigcitro: Thanks. Should the option be called `bigquery.quiet` (current state) or `bigrquery.quiet` (more consistent: <package.name>.<option name>)?

craigcitro:
Ah, sorry, missed that. Yes, I'd vote to switch over to `bigrquery.quiet`.

PR #78, opened 2016-01-25
DBI backend

### LINKED ISSUE #25: Provide DBI compatible interface
This should mostly involve providing methods for DBI generics. This will make dplyr implementation easy as it will be able to use more default methods.

krlmlr:
Fixes #25.

Closes #87 (=includes it).


R/dbi-connection.r
```diff
@@ -0,0 +1,169 @@
+#' @include dbi-driver.r
+NULL
+
+BigQueryConnection <- function(project) {
```
krlmlr: @hadley: I have two design questions here:

1. Is this a good name for the class's constructor?
2. Private slots seem to be [unsupported](http://stackoverflow.com/q/10755509/946850). What do you think about encapsulating the private slots in an environment? Would you choose a different name?

hadley: 1. Yes
1. I'd say using `.` as a prefix for private fields is fine by convention.

krlmlr: (2.) So, no separate environment, just private fields prefixed with a dot?

hadley:
LGTM

hadley:
When I run the tests locally I get a lot of `oauth_listener() needs an interactive environment.`. Any thoughts on how to get it to use my cached token?

krlmlr:
Wasn't this handled in #71? I can take a look tomorrow.

hadley:
Oh, that makes sense. I'll look into that too.

PR #81, opened 2016-01-26
New format_dataset() and format_table()

krlmlr:
as complement to parse_*().

@craigcitro: Okay to create a new file for these functions and move them out of utils.R? If not, I'll revert that particular change before adding tests.


R/id.R
```diff
@@ -0,0 +1,68 @@
...
+#'
+#' @param dataset dataset name
+#' @param project_id (optional) project ID to use if none is provided
+#' in \code{dataset}
+#' @return a list with \code{project_id} and \code{dataset_id} components
+#' (either of which may be \code{NULL}).
+#' @family identifier functions
+#' @export
+parse_dataset <- function(dataset, project_id = NULL) {
+  assert_that(is.string(dataset), is.null(project_id) || is.string(project_id))
```
craigcitro: small nit: i think this is easier to read when each assertion is on its own line, eg

    assert_that(is.string(dataset),
                is_null(project_id) || is.string(project_id))

R/id.R
```diff
@@ -0,0 +1,68 @@
...
+  first_split <- rsplit_one(dataset, ":")
+  dataset_id <- first_split$right
+  project_id <- first_split$left %||% project_id
+  list(project_id = project_id, dataset_id = dataset_id)
+}
+
+#' Format dataset and project ID as a BQ-style identifier
```
craigcitro: identifier.

R/id.R
```diff
@@ -0,0 +1,68 @@
...
+#'
+#' @param dataset dataset name
+#' @param project_id project ID
+#' @return a character.
+#' @family identifier functions
+#' @export
+format_dataset <- function(project_id, dataset) {
```
craigcitro: should we tweak the args here, so that

    do.call(format_dataset, parse_dataset(...)) == ...

R/id.R
```diff
@@ -0,0 +1,68 @@
...
+#' Parse a BQ-style identifier into project/dataset/table IDs.
+#'
+#' @param table table name
+#' @param project_id (optional) project ID to use if none is provided
+#' in \code{table}
+#' @return a list with \code{project_id}, \code{dataset_id}, and
+#' \code{table_id} components (any of which may be \code{NULL}).
+#' @family identifier functions
+#' @export
+parse_table <- function(table, project_id = NULL) {
```
craigcitro: it's strange that i apparently didn't add a `dataset_id` and use it if necessary. seems like it'd be worth adding?

craigcitro:
LGTM to me modulo comments above.

totally fine with moving these all out of `utils.R`.

R/id.R
krlmlr: I don't know, it's inconsistent all over the place: projectId vs. project vs. project_id. The first is required by the API, I guess, but we should be consistent at least throughout the bigrquery interface. I can change this instance, but there are a few others, most notably in jobs.R and tables.R.

R/id.R
krlmlr: There's usually no full stop in the caption of a function's documentation; I'll write a description for this function.

R/id.R
krlmlr: It's dataset, too. I'd like to keep the naming for this PR (so that it's consistent with the parse_() functions) and handle this issue elsewhere (#85).

R/id.R
krlmlr: #84

krlmlr:
Addressed your comments and added tests.

craigcitro:
LGTM, happy to resolve the other bits separately

PR #83, opened 2016-01-28
Removed unused argument.

mpancia:
There was a floating unused `callback` argument in `list_tabledata_iter()` that was causing my queries to throw an error.


krlmlr:
Good catch, thanks!

Issue #14, opened 2014-04-07
Figure out how to do automated testing with travis


hadley:

krlmlr:
Related: #22.

hadley:
Now seems like a matter of committing an encrypted service token.

hadley:
@craigcitro has the state of the art progressed here?

hadley:
Resolved 😄

PR #87, opened 2016-02-05
table data iter gains new members next_paged(), get_schema() and get_rows_fetched()

krlmlr:
Allows fetching all rows from a query, extracted DEFAULT_PAGE_SIZE as private variable.


R/tabledata.r
```diff
@@ -131,7 +133,16 @@ list_tabledata_iter <- function(project, dataset, table, table_info = NULL) {
     !is.null(last_response) && rows_fetched >= as.integer(last_response$totalRows)
   }
 
-  list(next_ = next_, is_complete = is_complete)
+  all <- function(page_size = DEFAULT_PAGE_SIZE) {
```
craigcitro: it might be worth taking this as an argument to `list_tabledata_iter`, even if just for testing purposes.

craigcitro:
One small nit, and the standard question -- can we add a test? :grin:

R/tabledata.r
krlmlr: Good point. I'll convert this to a functions, which then can be mocked by tests.

krlmlr:
#14.

krlmlr:
default_page_size() is now a function. Also added two more members.

R/tabledata.r
craigcitro: Ah, I actually meant something slightly different -- make this an explicit argument to `list_tabledata_iter`, and then pass an explicit value in tests. (it's also sometimes useful for end-users, though more rare.)

krlmlr:
re https://github.com/rstats-db/bigrquery/pull/87#discussion_r52267196: I think that it's a bit too much to support an argument to the outer function that specifies the *default* value (which can be overridden in all() anyway if necessary). For testing, default_page_size() can be mocked.

R/tabledata.r
```diff
@@ -52,7 +54,8 @@ list_tabledata <- function(project, dataset, table, page_size = 1e4,
 #' @export
 list_tabledata_callback <- function(project, dataset, table, callback,
                                     table_info = NULL,
-                                    page_size = 1e4, max_pages = 10,
+                                    page_size = default_page_size(),
```
hadley: I don't like using non-exported functions for defaults because it's hard to find out what the default value is

R/tabledata.r
```diff
@@ -131,7 +134,25 @@ list_tabledata_iter <- function(project, dataset, table, table_info = NULL) {
     !is.null(last_response) && rows_fetched >= as.integer(last_response$totalRows)
   }
 
-  list(next_ = next_, is_complete = is_complete)
+  all <- function(page_size = default_page_size()) {
+    ret <- list()
+    while (!is_complete()) {
```
hadley: This feels a bit icky, but I think doing it right will be a lot of work and is unlikely to have practical performance implication. Instead add a comment that says: "Fetching large data from bigrquery will be slow for other reasons..."

R/tabledata.r
krlmlr: I'll add an option instead.

krlmlr:
@craigcitro: The all() function is needed for DBI, and will be tested there implicitly. We probably want to set the default page size to something smallish there.

krlmlr:
Now using an option to define default page size, plus other enhancements. @craigcitro: PTAL.

hadley:
LGTM

craigcitro:
Just to check, is the diff here picking up anything extra? (In particular, the extra note at the top of NEWS looks like it should be on another PR?)

krlmlr:
The NEWS item was missing from another PR, I added it here for simplicity. Besides that, all changes relate to the list_tabledata_iter() function; I've also switched from integer to numeric to account for results with more than 2G rows.

PR #88, opened 2016-02-05
return empty data frame (instead of NULL) for zero-row result

krlmlr:
implemented by generating a one-row data frame full of NA values and then chopping off that one row


craigcitro:
LGTM -- is this easy enough to add a test for?

krlmlr:
Will be tested implicitly with one of the DBItest tests. I'm not looking into Travis testing just yet -- this requires setting up an OAuth token that can be used by Travis (#14).

krlmlr:
Merged with master.

R/tabledata.r
```diff
@@ -144,7 +144,13 @@ converter <- list(
 )
 
 extract_data <- function(rows, schema) {
-  if (is.null(rows)) return(NULL)
+  if (is.null(rows)) {
+    # Corner case: Zero rows
+    dummy_rows <- list(list(f = rep(list(NULL), length(schema$fields))))
+
+    data <- extract_data(dummy_rows, schema)
+    return(data[integer(), , drop = FALSE])
```
hadley: I think `data[0, , drop = FALSE]` is a more idiomatic way of doing this

PR #95, opened 2016-02-09
Use typeof() to detect data type

krlmlr:
instead of class().

From https://github.com/krlmlr/bigrquery/commit/50a1ff03afefa86dc7f7e4ef0bb0b8de294e5c81#commitcomment-15959160:

> this will give better default behaviour for base vectors with classes that we don't know how to deal with otherwise.


R/upload.r
```diff
@@ -71,15 +71,16 @@ schema_fields <- function(data) {
...
-    character = "STRING",
-    logical = "BOOLEAN",
-    numeric = "FLOAT",
-    integer = "INTEGER",
-    factor = "STRING",
-    Date = "TIMESTAMP",
-    POSIXct = "TIMESTAMP",
-    stop("Unknown class ", paste0(class(x), collapse = "/"))
+  if (is.factor(x)) return("STRING")
+  if (inherits(x, "POSIXct")) return("TIMESTAMP")
```
hadley: This should test for "POSIXt"

PR #102, opened 2016-02-21
Update copy_to.src_bigquery.

demorenoc:
Hi, basically passing the useful arguments (`create_disposition` and `write_disposition`) recently added to `insert_upload_job()` from the `copy_to.src_bigquery` method. Since this are very specific I think the method should have proper documentation which I added borrowing the relevant parts from `insert_upload_job()` and referencing it.

By the way, it would be nice to have a contributing guide.


krlmlr:
In `dplyr:::copy_to.src_sql()`, the copying fails if the table already exists. I'd expect the bigrquery implementation to behave identically, perhaps an `append_to()` generic could be added to dplyr?

demorenoc:
I see. That makes sense, but then that could be the default behaviour (which by the way I think is not the current behaviour in master either, that would require passing `write_disposition = "WRITE_EMPTY"` to the `insert_load_job()` and/or adding some code to manually check) so that it behaves exactly as `dplyr:::copy_to.src_sql()` **by default** but one could alter that with the parameters without having to use another method - I think that adding specific posibilities to a method for a given class is in the spirit of methods, for example note that the default in `dplyr:::copy_to.src_sql()` is to create tables as temporary (which can not be done with `insert_load_job()`) or for instance with `dplyr:::copy_to.src_sql()` you can pass indexes to be created which would not apply to BigQuery.

Still, that design choice is not mine to take and adding an `append_to()` (and even a `rewrite_to()`) generic could be nice and more general way to go in order to match the posibilities given by `insert_load_job()` in BigQuery to other databases, it is almost straight forward to build based on `copy_to()` given the generic db backend functions in particular `dplyr::db_insert_into()`.

krlmlr:
@demorenoc: Let's stick with `copy_to()`. I think overriding it (instead of implementing the functions that are called by `dplyr::copy_to.src_sql()`) is easier here, but I'd prefer if the code used `dbWriteTable()`. 

Also, it should have at least an `unique_indexes` argument and warn if it's set -- users might expect that this checks for some constraint, which (I think) is not possible with BigQuery.

hadley:
@demorenoc any chance you're still interested in this PR?

demorenoc:
@hadley not really, some comments seem relevant but I am currently not working with BigQuery. Thanks.

hadley:
No problems - I think this is mostly resolved with changes to dplyr anyway (since you can now specify `overwrite = TRUE` in `copy_to`)

Issue #58, opened 2015-07-30
list_tables() returns a maximum of 1000 table names


TroyHernandez:
`length(list_tables(project, dataset, max_results = 1001))`
`[1] 1000`

I have over 3000 tables in the dataset and could really use all of their names.

craigcitro:
so the BQ API has a limit on how many results it will return. 
* for bigrquery, we should definitely handle that paging ourselves -- i'm going to assume that's what you wanted here.
* that said, if you need something right now as a workaround, you could fish out the pagetoken and include it on the next request, and it'd let you pull the next 1000 tables. (and, of course, repeat as necessary)

from past experience, the slick thing to do here is to fetch lazily, which we should make sure we do.

TroyHernandez:
I'm not sure what `bq_get()` does, but it looks like it's getting more than just the names.  Right now I'm doing this as a workaround:

`length(system("bq ls --max_results=1001 project:dataset", intern = TRUE))`
`[1] 1001`

hadley:
@craigcitro do you know of any big public datasets that we could use for testing here?

craigcitro:
@hadley I think the easier approach is to expose the param that controls number of items per API call (either as a param or just for tests), and then just set that to 1 and try something with at least 2 items.

hadley:
Oh, duh, yeah.

hadley:
Duplicate of #108

TroyHernandez:
A little late to this... and it was two jobs ago, but IIRC bq_get() for this large number of tables was super slow because it was returning a bunch of additional unnecessary info... just a warning.

craigcitro:
Good call @TroyHernandez -- just filed #153 for that.

PR #103, opened 2016-02-21
Feat: list_tables() returns more than 1000 results

fieldryand:
Hi! This PR enables pagination through the results of the tables.list endpoint, allowing more than the default 1000 table names to be returned. This feature was requested in #58.

I see that `list_tabledata()` uses a callback to append the paginated results. I would be open to using the same approach here if you think it's beneficial.

Cheers!

### DIFF
--- a/R/tables.r
+++ b/R/tables.r
@@ -25,12 +25,34 @@ list_tables <- function(project, dataset, max_results = NULL) {
   if (!is.null(max_results)) {
     query$maxResults <- max_results
   }
-  data <- bq_get(url, query = query)$tables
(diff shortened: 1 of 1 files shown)


krlmlr:
I've noticed that max_results = NULL currently implies that 50 results are returned. Would you mind taking a look at this in the scope of this PR?

fieldryand:
Thanks for pointing that out--I will take a look.

hadley:
@fieldryand are you still interested in this PR?

hadley:
Actually I think it's better to do this at the request level so it's a bit more generalised

Issue #62, opened 2015-11-30
Unable to append data through query_exec


3vivekb:
I want to perform a query and grab a large amount of data from one table, then add it to another table that already contains data using R.

So if I select a table that already exists I get "Error: Already Exists: Table ic..."

In Bigquery what is happpening is we are 
1. Selecting a destination table
2. Writing if empty
While using the Bigquery Web console we can 
1. Selecting a destination table
2. Append to a table

Can we add this functionality to bigrquery's query_exec?  Is it available in the BigQuery API?  I imagine it gets complicated when the tables are different.

craigcitro:
the BQ API definitely supports this. we just need to thread the right arguments into `query_exec`.

3vivekb:
I think we should also have the feature to not return data.  R is a pretty convenient way to prototype and execute operations in order.

craigcitro:
sure -- that's just a call to `insert_query_job`, right?

3vivekb:
Oh ha.  Yea.  Hmm.  Maybe we should add all the features I suggested above to insert_query_job instead.

craigcitro:
I think we probably want them both places.

And a note in the docs that says "If you only need to run a query without retrieving results, see `insert_query_job`".

BrianWeinstein:
This would be super helpful.

As a (slow) workaround for the time being, you can download the results using `query_exec`, and then append to the desired table with `insert_upload_job`.

PR #111, opened 2016-03-08
insert_query_job(), query_exec(), and run_query_job() gain create_disposition and write_disposition args

BrianWeinstein:
Adds the functionality requested in #62. Allows the user to (1) control whether a job can create a new `destination_table`, and (2) specify the action that occurs if the `destination_table` already exists.

**create_disposition** specifies whether the job is allowed to create new tables
- "CREATE_IF_NEEDED" (default): creates the `destination_table` if it doesn't exist
- "CREATE_NEVER": returns an error if the `destination_table` doesn't yet exist

**write_disposition** specifies the action that occurs if the destination table already exists
- "WRITE_EMPTY" (default): returns an error if the `destination_table` already exists
- "WRITE_TRUNCATE": overwrite the `destination_table`
- "WRITE_APPEND": append results to `destination_table`


R/query.r
```diff
@@ -25,12 +25,17 @@
 #' query_exec(sql, project = project, default_dataset = "publicdata:samples")
 #' }
 query_exec <- function(query, project, destination_table = NULL,
-                       default_dataset = NULL, page_size = 1e4, max_pages = 10,
+                       default_dataset = NULL,
+                       create_disposition = "CREATE_IF_NEEDED",
```
craigcitro: any reason not to add these at the end?

craigcitro:
LGTM, just one little question above

BrianWeinstein:
Good catch. Just updated the argument order

PR #116, opened 2016-06-17
dplyr 0.5 compatibility [WIP]

hadley:
This switches to the new DBI based interface.  

I briefly explored making the package compatible with both dplyr 0.4 and dplyr 0.5, but it would be a lot of work. I think this approach is simpler.

PR #118, opened 2016-06-19
Review of dplyr 0.5 compatibility

krlmlr:
@hadley: If you merge this, it will update #116. I can also push directly to the other branch.

Did a brief test, a simple query worked for me. Noticed that copy_to() doesn't work yet. The error message mentions that dbBegin() isn't implemented. This means that the generic backend tests in dplyr won't work yet with bigrquery.


hadley:
Thanks!

Issue #117, opened 2016-06-19
Add possibility to execute query either with 'legacy SQL' or with new 'standard SQL'


mkazanov:
Google BigQuery recently added possibility to execute query in SQL 2011 standard:
https://cloud.google.com/bigquery/sql-reference/
It would be great to add, for example, a parameter to the query_exec function, which allows to switch from 'legacy SQL' to new 'standard SQL'

PR #124, opened 2016-07-01
Added support for sending the useLegacySql parameter in REST calls.

backlin:
As mentioned in issue #117, Google recently added support for executing queries written in [standard SQL](https://cloud.google.com/bigquery/sql-reference/) (in addition to legacy SQL). To use it in BigRQuery I just added an argument `useLegacySql` to a few functions and included the value in the REST call.

Thanks for a great package and hope you find this update useful!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bigrquery
 Title: An Interface to Google's 'BigQuery' 'API'
 Description: Easily talk to Google's 'BigQuery' database from R.
-Version: 0.3.1.9000
+Version: 0.3.0.9000
 Authors@R: c(
     person("Hadley", "Wickham", , "hadley@rstudio.com", c("aut", "cre")),
     person("RStudio", role = "cph")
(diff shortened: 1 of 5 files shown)


DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: bigrquery
 Title: An Interface to Google's 'BigQuery' 'API'
 Description: Easily talk to Google's 'BigQuery' database from R.
-Version: 0.3.0.9000
+Version: 0.3.1.9000
```
craigcitro: spurious?

craigcitro:
This looks good, modulo one little nit.

I suspect we're ultimately going to want a global setting for this, too, but this is a great start.

(And sorry about the slow review!)

DESCRIPTION
backlin: I wasn't sure how to deal with the version number, so feel free to change it to whatever you prefer. Because I had added functionality I though it was appropriate to at least increase the [patch number](http://semver.org/).

craigcitro: In general the version number only gets changed around releases; for Hadley's packages, to make development a little easier, he adds a `.9000` between last release and the next one.

Can you just revert this one hunk and we're good to go?

craigcitro:
Thanks!

PR #129, opened 2016-08-05
Add support for date type

### LINKED ISSUE #128: Error: Don't know how to convert type date
Querying a simple table containing a date types column with SELECT \* ... LIMIT 10 I get the error

```
library(bigrquery)
project <- "***" # put your project ID here
sql <- "select date_dt from test.table limit 10"
df <- query_exec(sql, project = project)

Error: Don't know how to convert type date
```

Selecting a subset of columns without the date column succeeds, and selecting only the date column generates the error.

kmatt:
Resolves #128

### DIFF
--- a/R/tabledata.r
+++ b/R/tabledata.r
@@ -169,7 +169,8 @@ converter <- list(
   float = as.double,
   boolean = as.logical,
   string = identity,
-  timestamp = function(x) as.POSIXct(as.integer(x), origin = "1970-01-01", tz = "UTC")
+  timestamp = function(x) as.POSIXct(as.integer(x), origin = "1970-01-01", tz = "UTC"),
+  date = as.Date
 )
 
 extract_data <- function(rows, schema) {
--- a/R/tabledata.r
+++ b/R/tabledata.r
@@ -170,7 +170,8 @@ converter <- list(
   boolean = as.logical,
   string = identity,
   timestamp = function(x) as.POSIXct(as.integer(x), origin = "1970-01-01", tz = "UTC"),
-  date = as.Date
+  date = as.Date,
+  bytes = as.raw
 )
 
 extract_data <- function(rows, schema) {


craigcitro:
So this seems totally reasonable to me -- @hadley, are there gotchas we should know about using `as.Date`?

hadley:
I don't remember how this code works, but the patch is unlikely to work given that `as.Date(1)` fails.

kmatt:
What's the proper approach, as this should be dealing with data enforced as proper dates from the database?

jarodmeng:
`Date` type is only recently added to Big Query in beta release of the standard sql (see [documentation](https://cloud.google.com/bigquery/sql-reference/data-types#date-type)). For a BQ table field to have the type `Date`, it follows the `'YYYY-[M]M-[D]D'` format which should be compatible with `as.Date()` in R. Since `converter()` will only be called on `Date` type field values, I don't see how it would run into calls like `as.Date(1)`.

craigcitro:
@hadley I think it's as safe as any of the other format conversions -- they all assume that the data coming in on the wire will be of the described format, which I think is safe?

Of course, a better first step would be to actually write some tests for `extract_data`, which can happen independent of getting any auth bits in place. 😉

page1:
This solved my ```Error: Don't know how to convert type date```

kmatt:
@hadley @jarodmeng And from that BigQuery doc page "Represents a logical calendar date. Values range between the years 1 and 9999, inclusive", this as.Date() should only receive valid date strings.  The code should not receive the infamous "0000-00-00" MySQL-valid date.

blakeyc:
Just checking if there is any updates/movement on this one, would be great to not have to remember to convert to timestamps where ever a date is used in queries.

inkrement:
Please add support for all the other Data Types (e.g., DateTime, Date and Time) too.

hadley:
Could someone please provide a minimal reprex that illustrates when this currently fails?

PR #130, opened 2016-08-05
Allow user to supply their own oauth app

jarodmeng:


jarodmeng:
Related to #130

craigcitro:
This seems totally reasonable -- @hadley may have feelings about option naming?

hadley:
I'd rather have one interface - i.e. options or `set_access_cred()`, not both.

jarodmeng:
@hadley I only went for the options route and didn't modify `set_access_cred()`. The option is for user to supply their own oauth app to be used in `oauth2.0_token()`.

hadley:
I get that - but why as an option instead of just exposing `set_access_cred()` so you can do whatever you want?

jarodmeng:
Mainly for convenience. To supply their own oauth app outside of `get_access_cred()`, user would need to 1) create an endpoint; 2) create an oauth app; 3) call `oauth2.0_token` with the endpoint, app and scope to get a token; 4) call `set_access_cred()` with the token. Using options inside `get_access_cred()` means user only needs to do 2) and put it in a global option.

hadley:
Or we could pull that bit out of `get_access_cred()` and provide a `set_oauth2.0_cred()` or similar with more parameters.

jarodmeng:
@hadley I added a `set_oauth2.0_cred()` which uses `bigqr` as default oauth app, but allows user to supply their own. `get_access_cred()` calls `set_oauth2.0_cred()` in the oauth dance.

hadley:
Are you still interested in this PR?  If so, I'd really appreciate a merge/rebase and a bullet in NEWS.

R/auth.r
```diff
@@ -28,16 +28,21 @@ bq_env <- new.env(parent = emptyenv())
...
+    set_oauth2.0_cred()
   }
 
-  cred
+  bq_env$access_cred
+}
+
+#' @rdname get_access_cred
+#' @export
+set_oauth2.0_cred <- function(app = bigqr) {
```
hadley: Oh `@app` needs documenting too. What sort of object should it be?

jarodmeng:
I vaguely recall something funky about the PR. I will take another look at the approach and rebase.

jarodmeng:
@hadley I rebased the PR on `rstats-db:master`, made some small changes and added a bullet to NEWS. I tried it on my machine using my own oauth app and it worked. LMK if there's anything I missed. Thanks!

NEWS.md
```diff
@@ -7,6 +7,8 @@
 
 * fixed for dplyr 0.6.0 compatibility
 
+* `set_oauth2.0_cred()` allows user to supply their own Google OAuth application when setting credentials (#130, @jarodmeng)
```
hadley: It doesn't matter here, but in the future, can you please add bullets at the top, not the bottom?

hadley:
Thanks!

PR #137, opened 2017-01-06
Added support for maximumBillingTier option closes #133

### LINKED ISSUE #133: Maximum tier
Is there any option to specify the maximum billing tier when querying ?

blakeyc:
#133 Includes the additional option of maximum_billing_tier when executing a query. Which defaults to NULL. Is only included in the request to API if not null.

This option will be useful for some of the upcoming work we will be doing.

### DIFF
--- a/R/query.r
+++ b/R/query.r
@@ -30,6 +30,7 @@ query_exec <- function(query, project, destination_table = NULL,
                        warn = TRUE,
                        create_disposition = "CREATE_IF_NEEDED",
                        write_disposition = "WRITE_EMPTY",
+                       maximum_billing_tier = NULL,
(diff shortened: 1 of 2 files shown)


R/query.r
```diff
@@ -48,13 +50,15 @@ query_exec <- function(query, project, destination_table = NULL,
 run_query_job <- function(query, project, destination_table, default_dataset,
                           create_disposition = "CREATE_IF_NEEDED",
                           write_disposition = "WRITE_EMPTY",
+                          maximum_billing_tier,
```
hadley: Should this default to `NULL`?

hadley:
Can you please also add a bullet to NEWS?

PR #139, opened 2017-01-14
Added query option flattenResults

emilyrbartha:
When using legacy SQL, all data saved to a table is flattened automatically. I added the query configuration option flattenResults so that the user can specify flattenResults = FALSE in order to preserve nesting and repeated fields for query results saved to a destination table.

### DIFF
--- a/R/jobs.r
+++ b/R/jobs.r
@@ -22,6 +22,7 @@
 #'   \code{query}, either as a string in the format used by BigQuery or as a
 #'   list with \code{project_id} and \code{dataset_id} entries
 #' @param useLegacySql (optional) set to \code{FALSE} to enable BigQuery's standard SQL.
+#' @param flattenResults (optional) set to \code{TRUE} to prevent BigQuery from automatically flattening nested and repeated data
 #' @family jobs
 #' @return a job resource list, as documented at
 #'   \url{https://developers.google.com/bigquery/docs/reference/v2/jobs}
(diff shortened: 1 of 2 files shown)


R/jobs.r
```diff
@@ -22,6 +22,7 @@
 #'   \code{query}, either as a string in the format used by BigQuery or as a
 #'   list with \code{project_id} and \code{dataset_id} entries
 #' @param useLegacySql (optional) set to \code{FALSE} to enable BigQuery's standard SQL.
+#' @param flattenResults (optional) set to \code{TRUE} to prevent BigQuery from automatically flattening nested and repeated data
```
hadley: Shouldn't you set it to `FALSE` to prevent flattening?

emilyrbartha: Whoops... yes, yes you should.

R/jobs.r
```diff
@@ -32,15 +33,17 @@ insert_query_job <- function(query, project, destination_table = NULL,
                              default_dataset = NULL,
                              create_disposition = "CREATE_IF_NEEDED",
                              write_disposition = "WRITE_EMPTY",
-                             useLegacySql = TRUE) {
+                             useLegacySql = TRUE,
+                             flattenResults = TRUE) {
```
hadley: This should be `flatten_results` (I'll fix `useLegacySql`) shortly.

R/jobs.r
hadley: Can you please fix, and wrap this long line.

R/query.r
```diff
@@ -48,14 +50,16 @@ query_exec <- function(query, project, destination_table = NULL,
 run_query_job <- function(query, project, destination_table, default_dataset,
                           create_disposition = "CREATE_IF_NEEDED",
                           write_disposition = "WRITE_EMPTY",
-                          useLegacySql = TRUE) {
+                          useLegacySql = TRUE,
+                          flattenResults = TRUE) {
   assert_that(is.string(query), is.string(project))
```
hadley: You don't seem to do anything with these new arguments?

hadley:
I'm going to close since you can now pass any extra arguments in `...`.  I'm going to look at the whole data retrieval process today, so it's likely I'll include this as part of that work.

PR #143, opened 2017-03-02
Dry run support

Zsedo:
Added support for dry run query to return query statistics. I found this extremely useful for large tables to get the cost of the query.


craigcitro:
Thanks for the PR! This is definitely something we'd like the support.

I think we may want to tweak the way we expose this: as it is, the two functions here (`query_exec` and `run_query_job`) now return different types based on one of their arguments, which is awkward.

WDYT of just adding a new `dry_run_query` function which takes the same args, and returns the query stats object? Users still have easy access, but now there's no worry about return types changing underneath you.

Zsedo:
@craigcitro I followed your example and did the dry_run_query as a standalone function. I agree that this is a much clearer solution.

DESCRIPTION
```diff
@@ -22,7 +22,7 @@ Suggests:
     withr
 License: GPL-3 | file LICENSE
 LazyData: true
-RoxygenNote: 6.0.1
+RoxygenNote: 5.0.1
```
byapparov: It seems that you changed and committed a lot of unrelated .Rd files because of this version rollback.

Zsedo:
@hadley Hello, sorry for the big delay on this. I've started working on this again yesterday in my hurry to integrate dplyr 0.7.0 changes and its communication with this package. I've made the recommended fixes. Please let me know what else there is to do.

craigcitro:
Could you squash and re-push?

Zsedo:
Could you help me out with this. My noobfu seems to have created a huge mess with squashing... Like all the stuff is there and I get what you want (to see the changes as a whole) but this is becoming quite tedious. I doubt that seeing all this commits from master branch now in this pull request was the intended result?

craigcitro:
Sure -- you've got a few choices. 

I'm about to list a bunch of git commands I haven't tried myself; as always, be cautious. 😉 I know just enough git-fu to get myself into trouble, so an easy strategy is to make a copy of the branch (`git checkout -b localcopy; git checkout master`) in case disaster strikes. 

I'm not sure what you called your remotes, so I'll use `github` for your github repo and `upstream` for this repo below. You can check with `git remote -v`.

1. The quick way out is as follows: rename this branch locally (`git branch -m master oldmaster`), and create a new master branch which is up to date with the upstream one (`git branch --track master upstream/master`). Then manually reapply changes (`git diff oldmaster | git apply`) or get crazy with `git rebase`. Finally, set the upstream to this branch (`git branch -u github/master`) and push (`git push -f`).

1. If you want to try recovering this branch, you'll want to do something like `git rebase --interactive upstream/master`, select the first commit and squash the rest, update the commit message and save, and then `git push -f`. This is much faster, but is a much larger gun you're potentially pointing at your foot. 😀

Zsedo:
Thanks @craigcitro that really helped out (did the 1.). :) Should be squashed now.

R/query.r
```diff
@@ -59,6 +59,22 @@ query_exec <- function(query, project,
   )
 }
 
+#' Submits a dry run query job, waits for it, and returns statistics of the dry run
+#' @export
+query_dry_run <- function(query, project, destination_table = NULL, default_dataset = NULL,
```
craigcitro: I think we should `...` all the args, rather than have them duplicated.

R/job-query.R
```diff
@@ -34,6 +35,7 @@ insert_query_job <- function(query, project,
                              create_disposition = "CREATE_IF_NEEDED",
                              write_disposition = "WRITE_EMPTY",
                              use_legacy_sql = TRUE,
+                             dry_run = FALSE,
```
craigcitro: Just to confirm: do we now pick this up for free, since `...` here is passed along to BQ?

byapparov: Current version of the code does not work because parameters are added to the root of the body instead of the `configuration` level. Here is PR https://github.com/r-dbi/bigrquery/pull/205 that fixes it and allows to pass any parameters (at configuration level including `dry_run` which will be correctly added as `dryRun`)

byapparov:
This is redundant after #205 is merged. `dryRun` will be added to the configuration of the request.

hadley:
Closing since it doesn't seem needed in concert with #205

Issue #174, opened 2017-08-04
`all()` translation to SQL


jarodmeng:
Currently `all()` is translated into `LOGICAL_ANY()` via [this snippet](https://github.com/rstats-db/bigrquery/blob/master/R/dplyr.r#L106) in `dplyr.r`.

```r
dbplyr::sql_translator(.parent = dbplyr::base_agg,
      ...
      all = sql_prefix("LOGICAL_ANY", 1)
    ),
```

The issue is that there's no `LOGICAL_ANY()` function in BigQuery. It should be [`LOGICAL_AND()`](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#logical_and).

PR #175, opened 2017-08-04
Fix all() SQL translation

jarodmeng:
Use `LOGICAL_AND()` instead of nonexistent `LOGICAL_ANY()`. Fixing #174.


hadley:
For future reference, if you include `Fixes #xyz` in a commit message (or the text of the PR), the corresponding issue will be automatically closed when the PR is merged.

Issue #176, opened 2017-08-04
Window function translation to SQL


jarodmeng:
Currently `sd()`, `all()` and `any()` window functions are translated to SQL using [the following snippet](https://github.com/rstats-db/bigrquery/blob/master/R/dplyr.r#L108).

```r
    dbplyr::sql_translator(
      .parent = dbplyr::base_win,
      sd = dbplyr::win_recycled("sd"),
      all = dbplyr::win_recycled("all"),
      any = dbplyr::win_recycled("any"),
      ...
    )
```

This produces wrong SQL translation as `sd()` should be translated to `STDDEV_SAMP()`; `all()` to `LOGICAL_AND()` and `any()` to `LOGICAL_OR()`.

Furthermore, BigQuery doesn't support `OVER()` clause for `LOGICAL_AND()` and `LOGICAL_OR()`. So `all()` and `any()` should be translated using `win_absent()` instead.

PR #177, opened 2017-08-04
Fix window function translations

jarodmeng:
Fixing #176.


hadley:
Thanks! 

If you have a minute (and you're still interested), would you mind writing up a bullet point to put in the NEWS.md describing all these improvements to SQL translation?

Issue #173, opened 2017-08-04
`%||%` translation to SQL


jarodmeng:
Currently `dplyr.r` has [this snippet](https://github.com/rstats-db/bigrquery/blob/master/R/dplyr.r#L100) to translate `%||%` function.

```r
dbplyr::sql_translator(.parent = dbplyr::base_agg,
      ...
      "%||%" = sql_prefix("concat"),
      ...)
```

This has two issues:

1. `%||%` should be a scalar function, not an aggregate function. Its parent should be `dbplyr::base_scalar`.
2. `%||%` shouldn't be translated into `concat()`. [`IFNULL()`](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#conditional-expressions) is probably a better translation.

The fix is rather straightforward. I can send out a PR to fix this if necessary.

PR #178, opened 2017-08-04
Fix %||% function's SQL translation

jarodmeng:
Fixing #173.


craigcitro:
LGTM -- but for both this and the others, is there a worthwhile test case to add?

jarodmeng:
@craigcitro PTAL and LMK if you want me to squash the two commits into one.

PR #184, opened 2017-08-25
only list_tabledata if max_pages*page_size > 0

ras44:
In the case users want to create a table via query_exec, but don't want to view any of its contents they might set page_size = 0 or max_pages = 0.  If max_pages = 0 an error is thrown during list_tabledata(...).  With this PR, list_tabledata is only called if the max_pages*page_size > 0.

### DIFF
--- a/R/query.r
+++ b/R/query.r
@@ -48,15 +48,17 @@ query_exec <- function(query, project,
     ...
   )
 
-  list_tabledata(
-    dest$projectId,
-    dest$datasetId,
-    dest$tableId,
-    page_size = page_size,
-    max_pages = max_pages,
-    warn = warn,
-    quiet = quiet
-  )
+  if(page_size*max_pages>0){
+    list_tabledata(
+      dest$projectId,
+      dest$datasetId,
+      dest$tableId,
+      page_size = page_size,
+      max_pages = max_pages,
+      warn = warn,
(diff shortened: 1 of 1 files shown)


R/query.r
```diff
@@ -48,15 +48,17 @@ query_exec <- function(query, project,
     ...
   )
 
-  list_tabledata(
```
byapparov: I think you should fix the `list_tabledata` function instead. 

e.g. assertion inside the function could be changed from:

```R
assert_that(is.numeric(max_pages), length(max_pages) == 1, max_pages >= 1)
```

to

```R
assert_that(is.numeric(max_pages), length(max_pages) == 1)
if (max_pages < 1) {
  return(data.frame())
}
```

Better alternative could be to get table schema with correct column names without getting the data.:

```R
assert_that(is.numeric(max_pages), length(max_pages) == 1)
if (max_pages < 1) {
  table_info <- table_info %||% get_table(project, dataset, table)
  extract_data(NULL, table_info$schema)
}
```

ras44:
@byapparov I added a new commit on a new branch that implements the second option: return table with fieldnames, but no rows.

byapparov:
@ras44 sounds good. I guess you can close this PR and create a new on in this case.

PR #155, opened 2017-04-25
Add functions to load CSV in Google Cloud Storage into a BQ table

jarodmeng:
Fix #126 

`insert_load_csv_from_cs_job()` inserts an [insert.load](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load) job to load a CSV file stored in Cloud Storage into BQ as a table. `load_csv_from_cs()` is a wrapper that waits for the insert job to finish.

PR #190, opened 2017-10-02
Add functions to load CSV in Google Cloud Storage into a BQ table (2nd attemp of #155)

### LINKED ISSUE #126: [FR] Function to load a table from Google Cloud Storage object(s)?
BQ allows [loading data directly from cloud storage](https://cloud.google.com/bigquery/docs/loading-data-cloud-storage). Currently `bigrquery` has a `insert_upload_job` function that writes a temp csv file from a data frame and loads the temp csv into a BQ table. Can we also have a function to load a table from one or many objects in Cloud Storage?  I think it can be done by using a simple `POST` to `https://www.googleapis.com/bigquery/v2/` with a `configuration.load` payload specifying the meta information. One extra bit we need to write to enable this might be to expand the scope of `get_access_cred` to allow accessing cloud storage.

Is someone already working on this feature? If not, I can take a stab and send out a pull request.

everron:
commits by @jarodmeng 

fixes #126


MiloParigi:
While we are waiting for Travis to be fixed, is this function reading data and creating a table in BigQuery or creating a permanent external table ? (cf https://cloud.google.com/bigquery/external-data-cloud-storage)

everron:
This creates a BigQuery table not an external table.

hadley:
Not sure how this is different to #155?

Issue #189, opened 2017-09-29
Authentication using Google's Metadata server when running in GCP


pcejrowski:
I run this library inside of Jupyter in Datalab in Google Cloud Platform. 
Currently, I use service account key that has to be stored inside of the Docker container. 
However, since it's run on Google Compute Engine, I have metadata server available and it should be possible to hide the authentication from a user and run it on the credentials provided by the metadata server.

MarkEdmondson1234:
You may be able to call this yourself using this code, adapted from [here](https://github.com/MarkEdmondson1234/googleAuthR/blob/master/R/gce.R#L28):

Thsi works on GCE VMs and App Engine, not tried it on Datalab yet though:

Run this on the server:

```r
gar_gce_auth <- function(service_account = "default",
                         client.id,
                         client.secret,
                         scope_list){
  
  call_url <- sprintf("http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/%s/token",
                      service_account)
  
  ## will be an error if not on GCE
  req <- try(httr::GET(call_url, httr::add_headers("Metadata-Flavor" = "Google")), silent = TRUE)
  
  if(is.error(req)){
    myMessage("Not detected as being on Google Compute Engine", level = 2)
    return(NULL)
  }
  
  token <- httr::content(req, type = "application/json")
  
  gar_app <- httr::oauth_app("google", key = client.id, secret = client.secret)
  
    httr::Token2.0$new(app = gar_app,
                       endpoint = httr::oauth_endpoints("google"),
                       credentials = list(access_token = token$access_token,
                                          token_type = token$token_type,
                                          expires_in = token$expires_in,
                                          refresh_token = NULL),
                       params = list(scope = scope_list, type = NULL,
                                     use_oob = FALSE, as_header = TRUE),
                       cache_path = "token.oauth")
  
}
```

hadley:
This will happen as part of #167

Issue #167, opened 2017-06-23
Use auth from gargle


hadley:
To support wider variety of methods.

jpvelez:
Hi there! Just wanted to check in on the status of this issue.

hadley:
And user agent helper #245

craigcitro:
Because some folks were asking: my understanding is that Jenny is tackling some googlesheets4 work; once that's done, she'll tackle gargle, and then integrate it into this library.

craigcitro:
Also, here "gargle" means https://github.com/r-lib/gargle

hadley:
Also library means package 😛

ZainRizvi:
Hi, is there any progress on this? I see there is an orphaned pull request by @jennybc from three months ago that would fix this issue.

Thanks

jennybc:
gargle has been submitted to CRAN 4 times in the last 10 days and keeps getting held up, despite passing all checks. Once this blocker is resolved, then, yes, gargle can be used for auth here and it should handle the GCE scenario.

ZainRizvi:
Gotcha, thanks for the context and thank you for making this much needed fix @jennybc! Looking forward to see it live :)

alokpattani:
Hi there, I work with @ZainRizvi at Google and wanted to +1 the desire for an improved auth experience in BigRQuery, which it looks like gargle will enable.

For greater context, I'm hoping to advocate for the use of R with Google Cloud Platform, and having an improved workflow for this piece will go a long way to that end. Looking forward to any updates you can share, and really appreciate the work!

jennybc:
CRAN continues to be the problem. I know of no obstacles at this point. gargle is just sitting there. 5 resubmissions so far and 3 weeks and lots of email 😬 I'm doing everything I know to do.

PR #195, opened 2018-01-15
Attempt automatic token generation using the google metaservice

haavardw:
The metadata service is available when running on a Google Compute Engine and provides
you with a compute instance service-account and credentials. Make an attempt on retrieving these credentials before doing the interactive flow.

As an example you're running in this environment when running R in a jupyter notebook with Google Datalab (https://cloud.google.com/datalab/). In this scenario the notebook http interface is proxied through an ssh tunnel and the interactive authentication fails. Generating and downloading the credentials as a keyfile is also not straight-forward in this case (the used might not have permissions to do this). Using the compute instance service-account credentials solves these issues.

Solves #189


hadley:
This will be handled at part of #167 so we can have the same code shared across multiple packages.

PR #198, opened 2018-02-01
Add names to schema from git_table

michaelquinn32:
This makes it easier to explore the schema in the Rstudio Viewer.

### DIFF
--- a/R/tables.r
+++ b/R/tables.r
@@ -70,7 +70,21 @@ get_table <- function(project, dataset, table) {
   assert_that(is.string(project), is.string(dataset), is.string(table))
 
   url <- sprintf("projects/%s/datasets/%s/tables/%s", project, dataset, table)
-  bq_get(url)
+  if (!is.null(metadata$schema)) {
+    metadata$schema <- name_recursively(metadata$schema)
+  }
+  metadata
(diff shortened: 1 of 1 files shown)


R/tables.r
```diff
@@ -70,7 +70,21 @@ get_table <- function(project, dataset, table) {
   assert_that(is.string(project), is.string(dataset), is.string(table))
 
   url <- sprintf("projects/%s/datasets/%s/tables/%s", project, dataset, table)
-  bq_get(url)
```
craigcitro: I suspect something was dropped here -- did you want this to be

    metadata <- bq_get(url)

instead?

michaelquinn32: Oops. It's put back.

byapparov:
Do you mind adding the tests?

Pull request is about get_table, not git_table 

It does not make sense to change the output of the ‘get_tables’ as one could use it to create similar tables (without tests its difficult to assert what are you trying to achieve).

If you want pretty output in console, there should be another function, e.g. summary() or print.schema() for that.

hadley:
I agree with @byapparov. Can you please start by filing an issue that describes the motivating problem?

PR #201, opened 2018-02-24
Translation

edgararuiz:
- Adds support for `paste0()`, `cor()` and `cov()`
- Updates NEWS


byapparov:
Should there be some tests for these features?

edgararuiz:
I couldn't find a test script with that tests other vector functions, so it felt out of scope to me.  The only one I found that was close was a type conversion test: https://github.com/r-dbi/bigrquery/blob/master/tests/testthat/test-sql.R

byapparov:
I think you are pointing to the right place. You should add tests for your code so reviewers could see what you are trying do achieve.

I think for this package tests should be mandatory on PRs as likely few people use it in production. I am not the maintainer and don't want to discourage you. If you need any help with tests let me know.

edgararuiz:
Thanks @byapparov ! You're right, we should have tests for this, and existing SQL vector function customization.

hadley:
@edgararuiz Do you want to merge/rebase?

I don't think it's necessary to test these translations.

edgararuiz:
Ok, I'll fix the conflicts

edgararuiz:
https://github.com/r-dbi/bigrquery/issues/216

PR #202, opened 2018-02-24
Adds DBI/dplyr refs to Readme

edgararuiz:
Brings to the front how to connect and use Big Query using `DBI` and `dplyr`


hadley:
Thanks!

PR #203, opened 2018-02-28
Added ability to create a table with a given schema.

byapparov:
This can be useful if schemas for the dataset are source controlled. For example CI jobs can create sample data using predefined schemas with correct types rather than relying on the implicit datatypes from data.frame upload.


byapparov:
implements #204

R/tables.r
```diff
@@ -29,11 +29,14 @@ list_tables <- function(project, dataset, page_size = 50, max_pages = Inf) {
 #'
 #' @inheritParams insert_dataset
 #' @inheritParams get_table
+#' @param schema schema of the table to be created
```
hadley: What is this supposed to contain?

R/tables.r
```diff
@@ -29,11 +29,14 @@ list_tables <- function(project, dataset, page_size = 50, max_pages = Inf) {
...
+#' @param partition adds time partitioning by DAY to the table with default column name (_PARTITIONTIME) if set to TRUE
```
hadley: I think it would be better to make this `"day"`, since the design of google api suggests that there might be other options in the future.

byapparov:
@hadley updated PR. not sure if that is exactly what you meant, but this should allow for further options for partition types in BQ.

hadley:
I took a different tack now that I've added a bunch of helpers around `bq_fields()`

PR #205, opened 2018-03-04
Fix dot dot dot

byapparov:
Fixes the way (`...`) parameters are added to the body of the request.


tests/testthat/test-request.r
```diff
@@ -22,3 +22,21 @@ test_that("explicit api keys override env vars", {
     Sys.unsetenv("BIGRQUERY_API_KEY")
   })
 })
+
+
+test_that("Dot-dot-dot parameter list is added to configuration params", {
+  body <- list(configuration = list(
+    query = list(query = "SELECT COUNT(*), 2 FROM crm.customers",
```
hadley: Would you mind using standard tidyverse style here? http://style.tidyverse.org/syntax.html#long-lines

I think you can also shorten the SQL since it's just a dummy value

R/request.r
```diff
@@ -35,7 +35,15 @@ bq_ua <- function() {
 
 bq_body <- function(body, ...) {
   user <- toCamel(list(...))
-  utils::modifyList(body, user)
+
+  if ("configuration" %in% names(body)) {
+    configuration <- utils::modifyList(body$configuration, user)
+    body$configuration <- configuration
+    return(body)
```
hadley: No need for an explicit return here

R/request.r
```diff
@@ -35,7 +35,15 @@ bq_ua <- function() {
...
+  }
```
hadley: `}` and `else` should come on same line

R/request.r
```diff
@@ -35,7 +35,15 @@ bq_ua <- function() {
 
 bq_body <- function(body, ...) {
   user <- toCamel(list(...))
-  utils::modifyList(body, user)
+
+  if ("configuration" %in% names(body)) {
```
hadley: A comment explaining why the behaviour is different would be useful

hadley:
Can you please add a bullet to NEWS? It should briefly describe the change (starting with name of the function), and crediting yourself with `(@yourname, #issuenumber)`.

R/request.r
byapparov: In cases with `if` it is a bit unclear what will be returned in that case.

or should I put:

```
body$configuration <- configuration
body
```

byapparov:
@hadley this PR is updated now.

NEWS.md
```diff
@@ -1,11 +1,18 @@
 # bigrquery 0.4.1.9000
 
+
+## New features
+
```
hadley: Can you please remove these headings?

tests/testthat/test-request.r
```diff
@@ -22,3 +22,28 @@ test_that("explicit api keys override env vars", {
...
+
+test_that("ellipsis parameter list is added to configuration params", {
+  body <- list(configuration = list(
+    query = list(query = "SELECT 1",
+                 useLegacySql = TRUE)
+  ))
+  res <- bq_body(body, dry_run = TRUE)
+  expect_equal(
+    res$configuration$dryRun, TRUE,
+    label = "camel case attribute is added to the body of the request"
```
hadley: Please don't use `label` in test

tests/testthat/test-request.r
```diff
@@ -22,3 +22,28 @@ test_that("explicit api keys override env vars", {
     Sys.unsetenv("BIGRQUERY_API_KEY")
   })
 })
+
+
+test_that("ellipsis parameter list is added to configuration params", {
+  body <- list(configuration = list(
```
hadley: Can you please follow the style guide here? You might want to use the styler package.

byapparov: I have made it more readable. Styler did not change this particular part, but it changed 22 files in the package! I think it will help if you run styler and merge changes so it was possible for other to use it.

byapparov:
@hadley I have done requested changes.

hadley:
I'm a bit confused here - it's not obvious what this is supposed to implement. Would you mind filing an issue describing the problem?  It's probably easier for me to implement given that I've completely changed the API.

byapparov:
Well, this was to support additional parameters that were passed via `...`. This functionality did not work as parameters were appended at the wrong level of the request body. 

DryRun was the original use case:
https://github.com/r-dbi/bigrquery/pull/143

hadley:
I see - I think to be consistent parameters in  `...` must always be added to the bottom-level - you can always embed in a list.

`dryRun` isn't particularly compelling because when `dryRun = TRUE` you also want to process the result in a different way. See `bq_query_fields()` for example.

PR #208, opened 2018-03-06
Fix copy function

byapparov:
There was a type which removed the body parameter and copy function throws and error:

> 
> Error in match(x, table, nomatch = 0L) : 
>   argument "body" is missing, with no default

### DIFF
--- a/R/tables.r
+++ b/R/tables.r
@@ -181,5 +181,5 @@ copy_table <- function(src, dest,
     )
   )
 
-  bq_post(url, body = bq_body(...))
+  bq_post(url, body = bq_body(body, ...))
 }


hadley:
Would you be interested in adding a unit test too?

byapparov:
`copy_table` function has tests. I think we need to fix the build first before we know what exactly needs more test.

hadley:
Yes, I'm working on that, but the tests obviously failed to pick this up before or I would've noticed such an obvious mistake.

byapparov:
Can you please confirm that all commits to master have passed the tests.

If they did I will have to do a bit of mocking to make a regression test for this specific bug.

hadley:
There's currently no real test for `copy_table()` - i.e. something that creates a table, copies it, and then checks the destination exists. That test is sufficiently simple that I think you can write it and check it independently of the other failing tests.

byapparov:
I have added a basic test within `dataset_10_tables()` dataset.

tests/testthat/test-tables.r
```diff
@@ -51,11 +51,21 @@ test_that("table references can be merged", {
   expect_that(merge_table_references(alternate, complete), equals(alternate))
 })
 
-test_that("copy_table validates its arguments", {
-  partial <- list(project_id = "a", dataset_id = "b")
-  complete <- list(project_id = "x", dataset_id = "y", table_id = "z")
+describe("copy_table", {
+  it("creates a copy of a table", {
```
hadley: Can you please stick to the testing style used elsewhere in the package? Mixing styles makes it difficult for me to maintain.

byapparov: np, done

PR #210, opened 2018-03-07
resolves #209: output reason in signal_reason

### LINKED ISSUE #209: responseTooLarge reason is not output in R/request.r::signal_reason()
When querying a table that contains many fields, I will occasionally receive the ambiguous output:

`Internal Error`

An example:
```
  SELECT *
      FROM `bigquery-public-data.samples.github_timeline`
```

The issue occurs in `R/request.r`: results are retrieved from `bq_get`, which calls httr `GET` and finally `process_request`.  The error that arises from `process_request` is passed to `R/request.r::signal_reason`, which signals a condition, but since this condition isn't handled, `signal_reason` ends with:
```
stop(message, call. = FALSE)
```

Unfortunately, the `message` output is uninformative, it only contains "Internal Error".  Helpful information is contained in `reason`.  In this case, it contains  `responseTooLarge`.

This specific issue can be solved by reducing the page_size argument, however my point is that `reason` contains additional information that is useful for troubleshooting.  Outputting `reason` and not just `message` in `R/request::signal_reason.r` would be helpful.

Another option is to handle the bigrquery_responseTooLarge class condition that is signaled, perhaps with a message recommending a smaller page_size argument.

ras44:


hadley:
I think this fix is in the wrong place - but I can handle it. Sorry I didn't merge your PR. I really appreciate the effort 😁

ras44:
No problem!  Thanks for handling it.

Issue #191, opened 2017-10-19
Add Support for Parameterized Queries


JHowix:
Via https://cloud.google.com/bigquery/docs/parameterized-queries#bigquery-query-params-api it is possible through the API to set and use parameters in a query.  It doesn't appear to be possible via the R package.

riccardopinosio:
I second that. I tried today but doesn't seem to be possible as yet. @hadley  is it difficult to implement? I might have a look at it myself at some point.

byapparov:
I see two options in terms of the how parameters argument can be structured:
```R 
# Explicit Type
query_exec(query = "SELECT * FROM orders 
                                    WHERE amount < @max_value AND first_name = @name ", 
                           project = "my-project",
                           parameters = list(max_value = list(type = "INTEGER", value = 10)))

# Implicit Type
query_exec(query = "SELECT * FROM orders 
                                    WHERE amount < @max_value AND first_name = @name ", 
                           project = "my-project",
                           parameters = list(max_value = 10, name = "Bob")))
```

It seems that other places in the package are relying on the implicit typing (which works most of the time). 

I will be happy to write the code for either version, should not be hard. Error handling is the only other unknown for this issue.

byapparov:
There is also an option to use positional parameters (`?`). lets say it is not part of this.

byapparov:
@riccardopinosio I have created PR, have a look.

PR #211, opened 2018-03-10
Added support for the named parameters

byapparov:
Implements #191

### DIFF
--- a/R/job-query.R
+++ b/R/job-query.R
@@ -22,7 +22,8 @@
 #'   `query`, either as a string in the format used by BigQuery or as a
 #'   list with `project_id` and `dataset_id` entries
 #' @param use_legacy_sql (optional) set to `FALSE` to enable BigQuery's standard SQL.
-#' @param parameters list of named parameters that will be used with parameterised SQL:
+#' @param parameters list of named parameters that will be matched to parameters in the query,
+#'    for example, `a` parameter will be matched to `@a` placeholder in the query.
 #'   \url{https://cloud.google.com/bigquery/docs/parameterized-queries#bigquery-query-params-api}
 #' @family jobs
 #' @return a job resource list, as documented at
--- a/R/job-query.R
+++ b/R/job-query.R
@@ -110,7 +110,7 @@ insert_query_job <- function(query, project,
 
 #' Converts named list to parameters data structure
 #' @noRd
-#' @param parameters named list of parameters for pamameterised query
+#' @param parameters named list of parameters for parameterised query
 bq_parameters <- function(parameters) {
   res <- lapply(names(parameters), function(name) {
     list(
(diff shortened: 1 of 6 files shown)


R/job-query.R
```diff
@@ -22,6 +22,8 @@
 #'   `query`, either as a string in the format used by BigQuery or as a
 #'   list with `project_id` and `dataset_id` entries
 #' @param use_legacy_sql (optional) set to `FALSE` to enable BigQuery's standard SQL.
+#' @param parameters list of named parameters that will be used with parameterised SQL:
```
hadley: Can you please describe what this should look like? An example would be helpful as well.

R/job-query.R
```diff
@@ -22,6 +22,8 @@
...
+#'   https://cloud.google.com/bigquery/docs/parameterized-queries#bigquery-query-params-api
```
hadley: This needs to be formatted as a link

R/job-query.R
```diff
@@ -82,3 +90,15 @@ insert_query_job <- function(query, project,
 
   bq_post(url, body = bq_body(body, ...))
 }
+
+
+bq_parameters <- function(parameters) {
+  res <- lapply(names(parameters), function(name) {
+    list(
```
hadley: I think it would be simpler to just provide a named character vector of big query field types, rather than forcing the user to create sentinel R objects.

tests/testthat/test-query-params.R
```diff
@@ -0,0 +1,52 @@
+context("query_exec parameters")
+
+
+test_that("named parameters can be used", {
```
hadley: Can you please fix the styling of this file?

R/job-query.R
```diff
@@ -47,6 +50,11 @@ insert_query_job <- function(query, project,
     )
   )
 
+  if (!is.null(parameters)) {
+    body$configuration$query$queryParameters <- bq_parameters(parameters)
+    body$configuration$query$parameterMode = "NAMED"
```
hadley: Should be `<-` not `=`

R/job-query.R
byapparov: This is internal function, user will need to pass a named list like this one:

```R
params <- list(x = 2L, d = as.Date("2012-10-15"))
```

Currently package is relying on the implicit type conversion, hence I followed that route.


If needed I am happy to change it if you can update this snippet:

```R
sql <- sql <- "SELECT *
          FROM m_table
          WHERE x = @x and date = @d"
params <- list(x = 2L, d = as.Date("2012-10-15"))
query_exec(
    sql,
    project = Sys.getenv("BIGQUERY_PROJECT"),
    use_legacy_sql = FALSE,
    parameters = params
)
```

byapparov:
@hadley this PR has been updated now.

R/job-query.R
hadley: I would expect `params <- c(x = "integer", d = "date")`

byapparov: where would the values go?

byapparov: @hadley I need more details to proceed with this. I am not sure how to fit in your suggestion in my example.

PR #213, opened 2018-03-15
Added function that allows to check whether credentials are set.

byapparov:
This can be useful in wrapper packages that only rely on service token (non-interactive server side authentication).

It will allow me to authenticate automatically without calling `set_service_token` function outside of the wrapper package code.


R/auth.r
```diff
@@ -26,14 +26,22 @@ bq_env <- new.env(parent = emptyenv())
...
   }
 
   bq_env$access_cred
 }
 
+#' Checks if access credentials are set
+#'
+#' @export
+#' @return TRUE if credentials are set
+is_set_access_cred <- function() {
```
hadley: Can you please call this `has_access_cred()` and document with `get_access_cred()`?

R/auth.r
```diff
@@ -26,14 +26,22 @@ bq_env <- new.env(parent = emptyenv())
...
+  cred <- bq_env$access_cred
```
hadley: I think you can make this one line

byapparov:
@hadley this PR is updated

R/auth.r
```diff
@@ -26,14 +26,20 @@ bq_env <- new.env(parent = emptyenv())
...
     set_oauth2.0_cred()
   }
-
   bq_env$access_cred
 }
 
+#' Checks if access credentials are set
+#'
+#' @export
+#' @return TRUE if credentials are set
```
hadley: Can you please `@rdname get_access_cred()`. I don't think it's worthwhile documenting such a small function separately.

byapparov: Done

byapparov: Yes, updated it to the way the documentation is done for `grep` like functions.

R/auth.r
```diff
@@ -25,15 +25,23 @@ bq_env <- new.env(parent = emptyenv())
...
-  cred <- bq_env$access_cred
-  if (is.null(cred)) {
+  if (!has_access_cred()) {
     set_oauth2.0_cred()
   }
-
   bq_env$access_cred
 }
 
+#' Checks if access credentials are set
```
hadley: You can remove this line and the subsequent line

R/auth.r
```diff
@@ -25,15 +25,23 @@ bq_env <- new.env(parent = emptyenv())
...
+#'
+#' @export
+#' @return has_access_cred returns TRUE if credentials are set
```
hadley: Our style is to put function names in backticks, and follow the name by parentheses, e.g. `` `has_access_cred()` ``. `TRUE` should also get backticks since it's R code.

byapparov: done

Issue #265, opened 2018-09-06
Two test fails with "Access Not Configured [...] JSON API."


rasmusab:
I've been trying, but failing, to get the test suit for `bigrquery` to run without tests failing. Two tests (`gs-object.R#27` and `test-gs-object.R#12`) fails due to the following reason:

> Access Not Configured. Cloud Storage JSON API has not been used in project 465736758727 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/storage-api.googleapis.com/overview?project=465736758727 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry. [accessNotConfigured]

If I click the link, I get to know that "you don't have permission", which makes sense as it's not my project.

**Could it be that enabling Cloud Storage JSON API for project 465736758727 would solve this?**

(Or could this be an issue on my side?)

How to replicate
----------------------------

Follow the instructions for testing found by running `?bq_test_init` in the R-prompt. Then run the following to run all the tests in the test suit:

```
library(testthat)
test_dir("tests/")
```

Full output when I run the test suite for `bigrquery`
------------------------

```R
> test_dir("tests/")


✔ | OK F W S | Context
── 1. Failure: can round trip data frame with list-cols (@test-bq-table.R#66)  ───
`df1` not equal to `df2`.
Incompatible type for column `val`: x numeric, y integer

── 2. Error: can roundtrip via save + load (@gs-object.R#27)  ───
Access Not Configured. Cloud Storage JSON API has not been used in project 465736758727 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/storage-api.googleapis.com/overview?project=465736758727 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry. [accessNotConfigured]
1: gs_object_delete(gs)
2: process_request(req) at .../coding/bigrquery/R/gs-object.R:27
3: bq_check_response(status, type, content) at .../coding/bigrquery/R/bq-request.R:171
4: signal_reason(json$error$errors[[1L]]$reason, json$error$message) at .../coding/bigrquery/R/bq-request.R:188

── 3. Failure: can retrieve full query results (@test-dbi-result.R#11)  ───
`df` not equal to tibble(count = 32).
Incompatible type for column `count`: x integer, y numeric

── 4. Failure: can retrieve without dataset (@test-dbi-result.R#17)  ───
`df` not equal to tibble(count = 32).
[...]

Issue #94, opened 2016-02-08
64-bit integers


krlmlr:
currently cannot be returned.

```
> query_exec("SELECT 10000000000", "bigrquery-1168")
  f0_
1  NA
Warning message:
NAs introduced by coercion to integer range 
> query_exec("SELECT 10000000000.1", "bigrquery-1168")
    f0_
1 1e+10
```

craigcitro:
yeah, this one is a serious bummer -- @hadley do you have strong opinions on any of the options out there for 64 bit integers in R?

krlmlr:
In DBI, 64-bit values will be represented as numeric, with an optional attribute that stores the original values as a character vector (see https://github.com/rstats-db/DBI/issues/42#issuecomment-171441544 and discussion above). Perhaps bigrquery could implement such a strategy right away, then there will be less trouble adapting that to DBI.

craigcitro:
Oh, that's definitely an easy way to go: in fact, for JSON upload/download, that's already how BQ sends data across the wire. :wink:

krlmlr:
Did you mean to comment in #97?

hadley:
We're standardising on bit64, but it's a hassle to return all integers as int64 because then most ordinary R code won't work.

hadley:
Maybe best thing is to have an option to return all integers at bit64s?

krlmlr:
I'm still not sure we should be returning raw `integer64` objects in DBI. I'd rather return something along the lines of

```r
structure(
  as.numeric(x),
  class = "xxx",
  integer64 = as.integer64(x)
)
```

with suitable `as.integer64.xxx()`, `as.character.xxx()`, and perhaps `[.xxx()` methods. This gives numerics by default but allows lossless retrieval of the original data.

Are you considering an option as an argument to `dbConnect()`?

hadley:
@krlmlr where did we end up API-wise on this with other packages? Argument to `dbConnect()`?

krlmlr:
Yes: https://github.com/r-dbi/DBItest/issues/133. Default would be `integer64`, but possible to downgrade to numeric or integer. Planned, not yet implemented.

randallhelms:
Just chiming in here - the Firebase connection to BigQuery stores event timestamp as a 64-bit integer. It's called event_dim.timestamp_micros, as you can see here:

https://support.google.com/firebase/answer/7029846?hl=en

[...]

PR #266, opened 2018-09-20
Add support for bigint argument to `dbConnect` and `bq_table_download`

rasmusab:
This pull request addresses issue #94 and tries to stay close to how 64-bit integers were handled in [RPostgres](https://github.com/r-dbi/RPostgres). That is:

* It adds support for setting the `bigint` argument to `dbConnect` and `bq_table_download` according to the DBI api: https://cran.r-project.org/web/packages/DBI/vignettes/spec.html#_specification_17
* It changes the default type of imported BigQuery `INT64` columns from R's `integer` to `bit64::integer64`.

Notes
---------------

* This is quite the breaking change as all integer columns will become `bit64::integer64` by default. But per [this discussion](https://github.com/r-dbi/bigrquery/issues/94#issuecomment-377065835) I've understood that this should be the default.
* I had to change a number of existing tests to work with `bit64::integer64` being the default return value.
* There were a couple of test that I never got to pass, but that I believe are unrelated to this pull request, which I have reported here: https://github.com/r-dbi/bigrquery/issues/265
* It's my obligation to state that this work was sponsored by my employer *King Entertainment* and that it was join work between me and @RickWalker .


hadley:
Looks good, but I think int64 by default is too much of a breaking change (at a minimum it means I'd have to think about it a lot which will delay a release). Can you please preserve the existing behaviour?

rasmusab:
The existing behavior is now preserved and the default is `bigint = "integer"`.

An added bonus (I think) is that as an intermediat step BigQuery's BigInt columns are still represented as `bit64::integer64` which means that overflows now results in a warning: 

``` R
con <- DBI::dbConnect(bigrquery::bigquery(), project = "my-google-project")
DBI::dbGetQuery(con, "SELECT 100, 100000000000")
```
```
# A tibble: 1 x 2                                                                                                                                                                   
    f0_   f1_
  <int> <int>
1   100    NA
Warning message:
In as.integer.integer64(X, ...) : NAs produced by integer overflow
```

My branch commit history is a bit messy, and while I've never messed around which changing commit history before, please let me know if I should tidy it up!

hadley:
Thanks @rasmusab!  I usually squash-and-merge, so no need to worry about a clean history.

PR #278, opened 2018-12-14
Typo fix for "write_disposition" parameter

414theodore:
Current documentation lists it as "write_dispoition". Without update, this throws a vague error, the root cause not easily discernible. I lost a couple hours before noticing the the parameter was misspelled in the documentation.


hadley:
Thanks for pointing this out - the fix actually needs to be made in the source `.R` file, so I made the fix myself.

PR #280, opened 2018-12-20
 Added auto_unbox to the JSON for the body of request.

byapparov:
This should prevent from this error when table is created with timePartitioning: 

>> Error: Invalid JSON payload received. Unknown name "type" at 'table.time_partitioning': Proto field is not repeating, cannot start list. [badRequest]


hadley:
This is a really wide ranging change, which makes me nervous. Is there a simpler way to fix the problem at hand?

hadley:
Oh I see the problem now — there's a few other bits that need changing, presumably because bigquery has become stricter with the JSON it accepts.

PR #292, opened 2019-02-26
Use gargle for auth

### LINKED ISSUE #167: Use auth from gargle

jennybc:
Closes #167 Use auth from gargle


R/bq-auth.R
```diff
@@ -0,0 +1,183 @@
...
+  }
+  .auth$set_cred(cred)
+
+  invisible()
+}
+
+#' @rdname deprecated-auth
+#' @keywords internal
+#' @export
+has_access_cred <- function() {
```
byapparov: @jennybc here is an example how we use it:

```r
bqAuth <- function() {
  if (!bigrquery::has_access_cred()) {
    bigrquery::set_service_token(bqTokenFile())
  }
}
```

this is useful in a wrapper package to avoid calling `set_service_token` multiple times in one process.

jennybc: Thanks.

The general m.o. of gargle is that a sequence of credential functions are called. At the first success, the token is cached for the rest of the session. So, with the caveat that we're both talking a bit in the abstract, you should be able to adapt pretty easily. For example, you could explicitly register your service token once to deal with this proactively.

But  I am very interested in hearing from users like you, in case we're at risk of taking a real step backwards anywhere.

jennybc: Now that gargle is on CRAN this PR is moving forward again 🎉

The most natural way to achieve what you do above now is:

``` r
bq_auth(path = bqTokenFile())
```

if you want to explicitly put a service account token in force.

Alternatively, if you ensure the JSON representation of the token is found in one of the places checked by Application Default Credentials, it will be discovered automatically:

https://gargle.r-lib.org/reference/credentials_app_default.html

---

Incidentally, I've come around to the fact that it is useful to have and export a function that reports whether we currently have a token. `has_access_cred()` is still present but it's deprecated, in favour of the new `bq_has_token()`, which is more consistent with the cross-package style we're trying to create with gargle.

jennybc:
I think the only thing missing now is describing this in the NEWS. But you could review now @Hadley.

R/bq-auth.R
```diff
@@ -0,0 +1,191 @@
+## This file is the interface between bigrquery and the
+## auth functionality in gargle.
+.auth <- gargle::init_AuthState(
+  package     = "bigrquery",
+  auth_active = TRUE
+)
+
+bq_app <- function() bqoa()
```
jennybc: Now handled like the credentials embedded in gargle.

I could even add the `check_permitted_package()` treatment?

R/bq-auth.R
```diff
@@ -0,0 +1,191 @@
...
+#'   bq_auth_config(app = google_app)
+#' }
+#'
+#' \dontrun{
+#' ## bring your own app via JSON downloaded from Google Developers Console
+#' bq_auth_config(
+#'   path = "/path/to/the/JSON/you/downloaded/from/google/dev/console.json"
+#' )
+#' }
+bq_auth_config <- function(app = NULL,
```
jennybc: Before I merge this, I will make this just a proper setter, since `bq_oauth_app()` exists for getting the current app.

jennybc: This is the vexing function I was referring to. We actually need to be able to specify `app = NULL` as a way of clearing the `.auth$app`.

R/bq-auth.R
```diff
@@ -0,0 +1,191 @@
...
+                    scopes = c(
+                      "https://www.googleapis.com/auth/bigquery",
+                      "https://www.googleapis.com/auth/cloud-platform"
+                    ),
+                    cache = gargle::gargle_oauth_cache(),
+                    use_oob = gargle::gargle_oob_default(),
+                    token = NULL) {
+  cred <- gargle::token_fetch(
+    scopes = scopes,
+    app = bq_oauth_app() %||% bq_app(),
```
hadley: Why doesn't this use `gargle_app()`?

R/bq-auth.R
hadley: Maybe call it `bq_auth_configure()` to make it more clear that it's a setter, not a getter?

R/bq-request.R
```diff
@@ -29,6 +29,8 @@ bq_path <- function(project, dataset = NULL, table = NULL, ...) {
 bq_ua <- function() {
   paste0(
     "bigrquery/", utils::packageVersion("bigrquery"), " ",
+    "(GPN:RStudio; )", " ",
```
hadley: Can you please remind me what this is?

R/old-auth.R
```diff
@@ -0,0 +1,76 @@
+# nocov start
```
hadley: Already covered by https://github.com/r-dbi/bigrquery/blob/master/.covrignore#L1

R/bq-auth.R
jennybc: Right now I'm just working with your existing credentials.

If we switched, it would be to `tidyverse_app()` btw. But there's no point until the verification situation becomes more clear.

R/bq-request.R
jennybc: This was the result of that meeting you had with someone at Google who wanted us to make the user agent compliant with some standard.

jennybc: The whole thing was rather anticlimactic, as I recall.

R/bq-auth.R
jennybc: It is unfortunately linked to its equivalent over in googledrive, which has long had `drive_auth_config()`. But I guess that's an easy change over there.

R/bq-request.R
jennybc: The only written record I have: https://github.com/r-lib/gargle/issues/37

jennybc:
> Would you mind also taking a look at `_pkgdown.yml`?

I don't see anything that needs to change there.

I think this is done. Can I merge @Hadley?

jennybc:
@byapparov I am merging this. If you get a chance to install the dev version soon from `master` and can take it for a test drive, I would be very grateful. We'll have a look around at other issues / PRs (I know some are yours), with an eye towards doing a release pretty soon.

byapparov:
@jennybc These are the two breaking changes from deprecated functions which are easy to fix:

`has_access_cred` -> `bq_has_token`
`set_service_token` -> `bq_auth`

Everything else seems to work on our tests, I am happy with this.

jennybc:
Thanks @byapparov and, to be clear / confirm, the original functions you use `has_access_cred()` and `set_service_token()` still work (right?) but you see a message alerting you to the switch.

byapparov:
Yes, they still work. Test that reference them fail because of the warnings.

Issue #272, opened 2018-10-09
Add description to bq_fields


KarimZaoui:
When creating table from schema, I would like to be able to specify a description for each field.

PR #295, opened 2019-03-03
Solves #272 - added description parameter to bq_field function.

byapparov:


tests/testthat/test-bq-table.R
```diff
@@ -134,3 +134,19 @@ test_that("can round trip data frame with list-cols", {
   df2 <- as.data.frame(df2)
   expect_equal(df1, df2)
 })
+
+test_that("can create table with time partitioning", {
```
hadley: Shouldn't this be a test for setting the description?

byapparov: I have updated the description of the test, not sure if you would prefer to have two separate tests for this.

hadley: I don't quite understand why time partitioning is in here at all? How is it related to the rest of this PR?

byapparov: You are right, this PR is not the best place for it. Hence removed the partitioning part from the test. 

At the same time I am very keen to have some regression testing for the ability to crate tables with partitions and to have this ability documented. Any suggestions?

PR #321, opened 2019-06-22
Added bq_table_stream function to stream data.frame into BigQuery table

byapparov:
bq_table_stream() is an alternative to inserting data into the table, which is covered by quota limits and will not work for a lot of small writes. 

This implementation does not split the data, which external implementation will need to manage. 

Recommended micro-batch size for BigQuery data streaming is 200-500 records per request, so something like this can be done to for larger datasets:

```r
my_table <- bq_table(project = "my-project", dataset = "test", table = "example")
df.list <- split(df, f = 1:nrow(df) %/% 200)
lapply(df.list, function(x) {
  bq_table_stream(my_table, x)
}
```
not sure if package should do this or delegate

Documentation for parameters is copied from here: 
https://github.com/googleapis/google-cloud-python/blob/0e231cab5f066ae5b3e254d0cb7d204e3dac7ebc/bigquery/google/cloud/bigquery/client.py#L1748

I could not find any documentation describing the fields in the request body on the API pages it self.


byapparov:
Replaced by #383

PR #327, opened 2019-06-28
Rebuild the test infrastructure

### LINKED ISSUE #326: Testing set up notes
I have shifted from the previous testing infrastructure (test project, test bucket, service account) and set it all up anew with our main GCP project.

This involved a few steps I had to discover through trial and error and I want to record a few notes for the future. This may not make sense to anyone but me.

  * Make sure target GCP project is associated with a billing account. I worked with RStudio devops to accomplish this.
  * In addition to BigQuery, the Cloud Storage JSON API must be enabled for the project. This was a mild surprise, but once I figured it out, I convinced myself it's necessary by turning in on, off, and on again.
  * The service account that is running the tests needs the BigQuery Admin and Storage Admin roles. Those might be overstated; I have not run an experiment with narrower roles within BigQuery and Storage, but definitely saw failure before I granted the service account these roles. Do this in the IAM console. You'll have to explicitly add the service account here, before you can assign it roles.
  * I set up the test bucket by hand in the browser and I'm not sure how @Hadley originally set his up. It does not seem to be done `bq_test_init()`? Or maybe when I first ran that, I had too many other problems for it to fully work? I still have questions about the test bucket, but I don't want to delete the current one, because we wouldn't be able to get the name back.

jennybc:
Notes for posterity are in #326. Closes #326.


.travis.yml
```diff
@@ -39,4 +39,3 @@ matrix:
 env:
   global:
   - _R_CHECK_FORCE_SUGGESTS_=false
-  - BIGQUERY_TEST_PROJECT=bigrquery-examples
```
jennybc: I've switched to a secure env var here.

DESCRIPTION
```diff
@@ -85,6 +85,5 @@ Collate:
     'old-query.R'
     'old-table.R'
     'old-tabledata.R'
-    'secret.R'
```
jennybc: We're using the next generation of these functions now, from their new home in gargle.

R/bq-test.R
```diff
@@ -1,21 +1,21 @@
 #' Project to use for testing bigrquery
 #'
-#' @description
-#' You'll need to set up the `BIGQUERY_TEST_PROJECT` (name of a project) and
-#' `BIGQUERY_TEST_BUCKET` (name of bucket) env vars in order to run bigrquery
-#' tests locally. I recommend creating a new project because the tests involve
-#' both reading and writing in BigQuery and CloudStorage.
+#' @description You'll need to set up the `BIGQUERY_TEST_PROJECT` (name of a
+#'   project) and `BIGQUERY_TEST_BUCKET` (name of bucket) env vars in order to
```
jennybc: I'm not clear on the role of the bucket. And I note that tests pass on travis although we do not set BIGQUERY_TEST_BUCKET env var there.

PR #330, opened 2019-06-30
Use C++11 and declare this constant to be unsigned long long

jennybc:
Eliminates these warnings (seen on win-builder and local VM):

```
BqField.cpp:39:5: warning: integer constant is too large for 'long' type [-Wlong-long]
     y = NA_INTEGER64;
     ^
BqField.cpp:186:7: warning: integer constant is too large for 'long' type [-Wlong-long]
       INTEGER64(x)[i] = v.IsString() ? parse_int64(v.GetString()) : NA_INTEGER64;
       ^
```


jennybc:
More context: I note that this header came in via PR https://github.com/r-dbi/bigrquery/pull/266 and draws heavily on RPostgres. And RPostgres has used C++11 (at least on Windows) since Dec 2017.

src/Makevars
```diff
@@ -1 +1,2 @@
 PKG_CPPFLAGS=-DSTRICT_R_HEADERS
+CXX_STD=CXX11
```
hadley: Do we need something in the DESCRIPTION too?

jennybc: I don't think so. In WRE:

> In order to use C++11 code in a package, the package’s Makevars file (or Makevars.win on Windows) should include the line
>
> `CXX_STD = CXX11`
> Compilation and linking will then be done with the C++11 compiler.
> 
> Packages without a src/Makevars or src/Makefile file may specify that they require C++11 for code in the src directory by including ‘C++11’ in the ‘SystemRequirements’ field of the DESCRIPTION file, e.g.
>
> `SystemRequirements: C++11`
> If a package does have a src/Makevars[.win] file then setting the make variable ‘CXX_STD’ is preferred, as it allows R CMD SHLIB to work correctly in the package’s src directory.

jennybc:
In favour of the status quo, @jeroen notes elsewhere that this warning does not appear with the new Windows toolchain.

hadley:
Closing since everything should be using the new windows tool chain now

jennybc:
Ah, the beauty of aging your PRs 😂

PR #353, opened 2019-10-22
Augment BigQuery error messages with client-specific advice.

deflaux:
There are several error conditions easily fixed by changing the value of the `bq_table_download` `page_size` parameter. This pull request augments with BigQuery error with some client-specific advice. It also repeats the advice via `message()` to help make it more visible.

```
# Table ID bigquery-public-data:irs_990.irs_990_2017
# Table Size 455 MB
# Number of Rows 300,910

bq_table_download(bq_project_query(
  BILLING_PROJECT_ID,
  query = 'SELECT * FROM `bigquery-public-data.irs_990.irs_990_2017`'),
  bigint = 'integer64',
  page_size = 1000)  # <-------- Succeeds!
# page_size = 50000) # <-------- Fails with 'responseTooLarge' error.
```

```
# Table ID bigquery-public-data:github_repos.sample_commits
# Table Size 2.49 GB
# Number of Rows 672,309

bq_table_download(bq_project_query(
  BILLING_PROJECT_ID,
  query = 'SELECT * FROM `bigquery-public-data.github_repos.sample_commits`'),
  page_size = 100000) # <-------- Succeeds!
# page_size = 1000)   # <-------- Fails with 'rateLimitExceeded' when run in an
                      #           environment with a fast network such as AI Platform notebooks.
```


hadley:
Thanks!

PR #367, opened 2020-01-23
Added release of temporary files

spgarbet:
The library uses temporary files in a DBI::dbGetQuery that it never releases. This has caused an app to run out of '/tmp' space on my cluster and crash. I'm using this as a patch that has thus far prevented the problem.


hadley:
Duplicate of #371 (which also deletes another file). Thanks for contributing the code!

Issue #343, opened 2019-08-12
File error (Out of disk space) during `bq_table_download`


mr2dark:
I would like to run an R script which uses bigrquery to run a query to BQ and download results within an AWS Fargate task which is rather disk space constrained. There are a lot of data and when I run that script it eventually fails with the following lines in logs:
```
Error in file(path, open = "wb") : cannot open the connection
Calls: <Anonymous> ... bq_download_pages -> <Anonymous> -> <Anonymous> -> file
In addition: There were 50 or more warnings (use warnings() to see the first 50)
```

I assume that the cause of the problem is in the fact that `bq_download_pages` creates a set of temporary files which aren't removed after being parsed.
As a workaround (rather nasty) I do the following after a call to `bq_download_pages`:
```r
# A workaround cleanup until https://github.com/r-dbi/bigrquery/issues/343 is fixed
do.call(file.remove, list(list.files(tempdir(), pattern = 'bq-[[:xdigit:]]+\\.json', full.names = TRUE)))
```

But I believe the proper solution will be adding before the final line of `bq_download_pages` the following line:
```r
do.call(file.remove, page_paths)
```

meztez:
The file haven't been parsed at this point, files are parsed inside bq_table_download, one solution would be to add it after
https://github.com/r-dbi/bigrquery/blob/5f4322866780badd5aec58e0e16199a7844c1ba5/R/bq-download.R#L89

hadley:
Fixed by #372

PR #372, opened 2020-03-11
Clean tmp files after bq_table_download

meztez:
I don't see any particular reason to keep the files after they have been parsed. What do you think?

In reference to https://github.com/r-dbi/bigrquery/issues/343

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,7 +19,7 @@ Where to learn more:
 
 ### Changes that a user will notice
 
-Temporary files are now deleted after table download. (@meztze, #343)
+Temporary files are now deleted after table download. (@meztez, #343)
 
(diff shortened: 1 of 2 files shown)


hadley:
Looks good :smile: Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

NEWS.md
```diff
@@ -19,6 +19,8 @@ Where to learn more:
 
 ### Changes that a user will notice
 
+Temporary files are now deleted after table download. (@meztze, #343)
```
hadley, line 22: Can you please put it under "development version"? (At the top of the file)

hadley:
Thanks!

Issue #313, opened 2019-06-06
'Unknown type GEOGRAPHY' when fetching GIS data.


stelsemeyer:
I have trouble fetching geographical data from BigQuery, produced by functions like `ST_GEOGFROMTEXT`:

```
library(bigrquery)
billing <- bq_test_project()
sql <- "SELECT ST_GEOGFROMTEXT('POINT(15 20)') AS geog"
tb <- bq_project_query(billing, sql)
data <- bq_table_download(tb)
```
errors out with
```
Error in bq_parse_files(schema_path, page_paths, n = page_info$n_rows,  : 
Unknown type GEOGRAPHY
````
<sup>(unfortunately I could not get this code rendered with reprex.)</sup>

Is there any way to coerce those data types to strings/character by default? I saw that `readr` is doing some type detection in the background.

**Comment by geoffreylarnold** [comment]
So the way I've handled this is done the SQL differently and then converted to an sf object once the data is in R.

```
library(bigrquery)
library(sf)
billing <- bq_test_project()
sql <- "SELECT ST_ASTEXT(ST_GEOGFROMTEXT('POINT(15 20)')) AS geog"
tb <- bq_project_query(billing, sql)
data <- bq_table_download(tb) %>%  
     st_as_sf(wkt = "geog")
```

Tartomas:
Hello, I have a similar issue and I would like to resolve it using dplyr in order to get the show_quey() and then translate it to SQL. But when I can't get rid of the "geometry" field neither call it to my environment like this: 

`
data = tbl(con, "data") %>% 
    select(-location)
data
Complete
Billed: 0 B
Downloading 11 rows in 1 pages.
Error in bq_parse_files(schema_path, page_paths, n = page_info$n_rows,  : 
  Unknown type GEOGRAPHY
`
Do you know how to resolve this? 
Thanks!

denironyx:
Hey @Tartomas, did you find a solution to the comment you made?? I am kinda in a similar pitfall at the moment.

hadley:
bigrquery doesn't currently support GEOMETRY columns, sorry.

chrowe:
> bigrquery doesn't currently support GEOMETRY columns, sorry.

Has this changed or is https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#geography_type something different?

denironyx:
Nope, it is still the same.

PR #376, opened 2020-04-02
fix geography type parse error

kevinwang09:
Relevant issue: https://github.com/r-dbi/bigrquery/issues/313
Description: Whenever the type of field is GEOGRAPHY, it will fail to be parsed into R. 
Fix: add one extra exception (in src/BqField.cpp) to parse GEOGRAPHY into characters in R. 
For discussion: will it make more sense to parse on all other unexpected field types as characters into R?

Example code
```
library(DBI)
library(bigrquery)

con <- dbConnect(
  drv = bigrquery::bigquery(),
  project = "bigquery-public-data",
  dataset = "covid19_jhu_csse",
  billing = billing ## Whatever your billing is
)

dbListTables(con) ## List all tables in this connection

bigrquery::bq_table_fields("bigquery-public-data.covid19_jhu_csse.summary") ## List field types 

DBI::dbGetQuery(con, "SELECT *
FROM `bigquery-public-data.covid19_jhu_csse.summary`
LIMIT 10;")
```


denironyx:
Hi @kevinwang09, what's the aim of this issue?? I think it will be a great improvement if R can support the geography data type. Instead of first converting the geography values to strings using the ST_ASTEXT or ST_ASGEOJSON function.

hadley:
I'm with @denironyx here; I'd we should either create high fidelity columns or assume the user will handle it at the SQL level.

edzer:
A candidate native R representation would be an `sfc` column, possibly with `EPSG:4326` as default coordinate reference system, but hopefully with properly set CRS (=datum) information. This would then work with package `sf` and `ggplot2::geom_sf()`.

edzer:
I think both @paleolimbot and @etiennebr have experience with interacting with Big Query GIS from R.

denironyx:
Hi @edzer , 

This was my workaround. 

`### Extracting using the sql and bigrquery
query <- "SELECT * EXCEPT(wkt), ST_ASTEXT(wkt) AS Geometry FROM od_matrix.cartographic_partitions"
harambee_tb <- bq_project_query("harambee-prod", query)
data_haram <- bq_table_download(harambee_tb) %>% 
  sf::st_as_sf(wkt = "Geometry")`

But, they should be a better way to address this issue in the future. I think.

Issue #375, opened 2020-03-25
Redundant notFound exception when calling DBI::dbExecute("drop table ..")


valentas-kurauskas:
When calling `drop table`, `drop view`, etc commands for an *existing* table (view, etc) through `DBI::dbExecute`, the command is completed, but an exception is thrown. This makes it necessary to surround each such query with `try()` as a workaround.

---

```r
con <- DBI::dbConnect(bigrquery::bigquery(), project="myProject", dataset="myDataset")
#priority parameter is ignored
DBI::dbExecute(con, "drop table myDataset.myTable")
#Error : Not found: Table myDataset.myTable [notFound]
DBI::dbDisconnect(con)
```

meztez:
After execution, it is trying to retrieve the numbers of rows from the table listed in the job meta.

From https://github.com/r-dbi/bigrquery/blob/ff5c22e3aba28dc586b3805d11bc69c1da9b9dd6/R/dbi-result.R#L11

```
$configuration$query$destinationTable
$configuration$query$destinationTable$projectId
[1] "projectId"

$configuration$query$destinationTable$datasetId
[1] "datasetId"

$configuration$query$destinationTable$tableId
[1] "tableId"
...
$statistics$query$statementType
[1] "DROP_TABLE"

$statistics$query$ddlOperationPerformed
[1] "DROP"
```

I see a couple ways to avoid this error, the more obvious is to condition on the statementType.

But you could always use the more straight forward method DBI::dbRemoveTable.

hadley:
Minimal reprex:

``` r
library(bigrquery)

bq_auth(email = 'hadley@rstudio.com')

con <- DBI::dbConnect(bigrquery::bigquery(), project = bq_test_project())
DBI::dbExecute(con, "drop table myDataset.myTable")
#> Error: Not found: Dataset gargle-169921:myDataset was not found in location US [notFound]
#Error : Not found: Table myDataset.myTable [notFound]
DBI::dbDisconnect(con)
```

<sup>Created on 2020-05-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley:
Better reprex:

``` r
library(bigrquery)

ds <- bq_test_dataset()
con <- DBI::dbConnect(bigrquery::bigquery(), 
  project = ds$project,
  dataset = ds$dataset
)

DBI::dbExecute(con, "DROP TABLE IF EXISTS foo")
[...]

PR #378, opened 2020-05-01
extracting nrow from dropped table

meztez:
from #375,

I used an attributes, might be other ways to accomplish the same thing. What do you think?


hadley:
I tackled this slightly higher up, but thanks for this code — it made it much faster for me to come up with a solution.

meztez:
nice a bunch of people will be happy to be dropping table left and right, thanks

PR #382, opened 2020-05-04
Added .Deprecated() call inside all exported old functions

### LINKED ISSUE #335: Drop old files and functions
As version has progressed beyond `1.1.0` it would make sense to drop all the old code that comes from prior `1.0.0` to make it cleaner, e.g. `old-*.R` files and corresponding tests. 

I don't think anyone using  the old API should be able to notice the difference between `1.0.0` and `1.1.0`.

byapparov:
Old functions can be removed in the consequent release.

I have also removed the tests for the deprecated functions as they can be changed.

Closes #335


hadley:
Thanks! Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

byapparov:
@hadley, I have added a line for this change in News.md

PR #383, opened 2020-05-04
bq_stream() function to stream data into tables

byapparov:
bq_table_stream() is an alternative to inserting data into the table, which is covered by quota limits and will not work for a lot of small writes. 

This implementation does not split the data, which external implementation will need to manage. 

Recommended micro-batch size for BigQuery data streaming is 200-500 records per request, so something like this can be done to for larger datasets:

```r
my_table <- bq_table(project = "my-project", dataset = "test", table = "example")
df.list <- split(df, f = 1:nrow(df) %/% 200)
lapply(df.list, function(x) {
  bq_table_stream(my_table, x)
}
```
not sure if package should do this or delegate

Documentation for parameters is copied from here: 
https://github.com/googleapis/google-cloud-python/blob/0e231cab5f066ae5b3e254d0cb7d204e3dac7ebc/bigquery/google/cloud/bigquery/client.py#L1748

I could not find any documentation describing the fields in the request body on the API pages it self.


byapparov:
@hadley I have replaced bq_stream() PR with this clean one from upstream.

R/bq-stream.R
```diff
@@ -0,0 +1,89 @@
+#' Streams data from data.from into BigQUery
```
hadley, line 1:
```suggestion
#' Stream data.frame into into BigQUery
```

Can you please also add a sentence or two about why you'd use this function? (i.e. what you have in the description of this PR)

R/bq-stream.R
```diff
@@ -0,0 +1,89 @@
...
+#'   treats unknown values as errors.
+#' @param template_suffix
+#'   (optional) treat `name` as a template table and provide a suffix.
+#'   BigQuery will create the table `<name> + <template_suffix>` based
+#'   on the schema of the template table. See
+#'   https://cloud.google.com/bigquery/streaming-data-into-bigquery#template-tables
+#' @return NULL if there are no errors, otherwise list of errors with reasons
+#' @export
+#' @rdname bq-streaming
+#' @seealso https://cloud.google.com/bigquery/streaming-data-into-bigquery
```
hadley, line 24: Can you please include an example wrapped in `\dontrun{}` that shows how you'd use a for loop to send stream a big data frame?

R/bq-stream.R
```diff
@@ -0,0 +1,89 @@
...
+bq_table_stream <- function(x,
+                            values,
+                            insert_ids = bq_stream_random_id(nrow(values)),
```
hadley, line 27: I think it's dangerous to supply this; I think it would be safer for the user to always supply.

R/bq-stream.R
```diff
@@ -0,0 +1,89 @@
...
+                            skip_invalid_rows = "false",
+                            ignore_unknown_values = "false",
```
hadley, line 29: Shouldn't these be `FALSE`?

R/bq-stream.R
```diff
@@ -0,0 +1,89 @@
...
+                            template_suffix = NULL) {
+  x <- as_bq_table(x)
+  url <- bq_path(x$project, x$dataset, x$table)
+  url <- paste0(url, "/", "insertAll")
```
hadley, line 33:
```suggestion
  url <- bq_path(x$project, x$dataset, x$table, "insertAll")
```

R/bq-stream.R
```diff
@@ -0,0 +1,89 @@
...
+  body <- bq_stream_body(
+    x = values,
+    insert_ids = insert_ids,
+    skip_invalid_rows = skip_invalid_rows,
+    ignore_unknown_values = ignore_unknown_values,
+    template_suffix = template_suffix
+  )
+
+  response <- bq_post(url, body)
+  invisible(response$insertErrors)
```
hadley, line 43: What's the motivation for returning errors, rather than throwing an error that the user has to deal with?

R/bq-stream.R
```diff
@@ -0,0 +1,89 @@
...
+
+  body <- list(
+    kind = "bigquery#tableDataInsertAllRequest",
+    skipInvalidRows = skip_invalid_rows,
+    ignoreUnknownValues = ignore_unknown_values,
+    rows = rows
+  )
+
+  if (!is.null(template_suffix)) {
+    body["templateSuffix"] <- template_suffix
```
hadley, line 74:
```suggestion
    body$templateSuffix <- template_suffix
```

R/bq-stream.R
byapparov: User might not want to fail if some of the records were not injected.

R/bq-stream.R
byapparov: Sometimes you might have a data stream that does not have assigned IDs for measurements. 

These IDs control de-duplication process on BigQuery side, so giving random unique IDs should be safe.

hadley: It's not safe if the stream fails part way (and isn't that the point of streaming?). I think if you want to stream you need to first give a unique id to each row. If you're uploading in batches, and have to repeat part of a batch, a random id isn't going to help.

R/bq-stream.R
hadley: Then I think you should be returning visibly, and it needs to be clear what the user needs to do to with this data.

byapparov:
@hadley I will reply with changes to your review later. I don't think this is essential for upcoming release.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # bigrquery (development version)
 
+* `bq_stream()` allows streaming data into BigQuery table.
```
byapparov, line 3:
```suggestion
* `bq_stream()` allows streaming data into BigQuery table (@byapparov, #383)
```

hadley:
Looks like this is now a legacy API, so probably not worth merging at this point. Again, apologies, for taking so long to get back to you on this.

PR #393, opened 2020-06-30
Add minimal support for GEOGRAPHY field type

### LINKED ISSUE #388: BiqQuery fails on type "GEOGRAPHY"
It looks like geography columns can't be interpreted when they come through the query:

``` r
library(bigrquery)
library(DBI)

billing <- "optimum-door-109617"
con <- dbConnect(
  bigrquery::bigquery(),
  project = "publicdata",
  dataset = "samples",
  billing = billing
)

dbGetQuery(
  con, 
  "SELECT ST_GEOGFROMTEXT('POINT (30 10)') as col", 
  n = 10
)
#> Error in bq_parse_files(schema_path, page_paths, n = page_info$n_rows, : Unknown type GEOGRAPHY
```

In the JSON these come through as well-known text strings that could be passed on in the same way as a STRING type (or could theoretically use the [wk_wkt](https://paleolimbot.github.io/wk/reference/wkt.html) class, with the full disclaimer that I wrote that class).

It's worth noting that there is a fairly easy workaround (use `ST_ASTEXT()` around a geography column).

Happy to PR if this is a feature that would be welcome!

paleolimbot:
This is a PR to fix #388, and addresses issues identified with the previous PR (#376) that was closed.

The motivation is to make it possible for another package to wrap the [extensive BigQuery geography API](https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions), making use of the high-performance parsing and authentication management in this package. While it's possible to for a *user* to provide a workaround (e.g., using `ST_ASTEXT()`), it is difficult for a *package* to provide a workaround, since it doesn't know which (if any) columns contain geography that should be converted (very possible I'm missing an obvious workaround).

Options (as I see them) are:

- Continue to error on type GEOGRAPHY and make extension packages parse the JSON themselves if they want to support this column type (realistically, this would mean other packages would just copy `BqFields.cpp` and adapt for their own purposes)
- Return GEOGRAPHY as a character vector (this was the approach of #376)
- Return GEOGRAPHY as a character vector with a dummy class (this is the approach RPostgreSQL takes with unknown column types, where PostGIS columns are coerced to hex binary and given the class "pg_geometry")
- Return GEOGRAPHY as a [wk::wkt()](https://paleolimbot.github.io/wk/reference/wkt.html) vector, which is a thin wrapper around a character vector with a vctrs implementation (I maintain the wk package)
- Parse GEOGRAPHY into an [s2_geography vector](https://r-spatial.github.io/s2/reference/as_s2_geography.html), which uses the same C++ library as BigQuery geography (myself and @edzer maintain the s2 package)
- Parse GEOGRAPHY into an `sfc` from the sf package (maintained by  @edzer)

In this PR I took a middle-of-the-road approach with using `wk::wkt()`: the `wkt()` class doesn't do any parsing, but provides a recognizable class on which other packages can build without knowing anything about `bigrquery`. I'm happy to implement any of these, but I don't think continuing to error or returning a plain character vector is the way  to go.

I'm curious what @etiennebr and @edzer think about this, as they have practical experience with SQL and GIS that I don't.


edzer:
I think that if you can go for binary or hex binary transfer, you should always do that.

edzer:
`ST_ASBINARY` gives WKB, but indeed as `BYTE`; as long as `bigrquery` doesn't support `BYTE`, your current WKT proposal seems the best possible.

hadley:
@paleolimbot if you're game to implement `BYTE` support, I'm happy to provide any help I can (with the caveat that I don't have bigrquery loaded in my head at the moment so you probably know more about the problem than me)

hadley:
Sorry I missed this, my preference would be to return a blob since that's most similar to other databases, and where we're heading in general.

hadley:
Overall, looks good to me, but I'll get someone else to check out the C/C++ code since I'm probably currently blind to any major issues.

src/BqField.cpp
```diff
@@ -253,6 +270,30 @@ class BqField {
     case BQ_RECORD:
       SET_VECTOR_ELT(x, i, recordValue(v));
       break;
+    case BQ_GEOGRAPHY:
+      if (v.IsString()) {
+        Rcpp::RObject chr = Rf_mkCharLenCE(v.GetString(), v.GetStringLength(), CE_UTF8);
```
DavisVaughan, line 275: I'm not sure if this matters that much, but I think CHARSXP maps to a Rcpp String object?

```suggestion
        Rcpp::String chr = Rf_mkCharLenCE(v.GetString(), v.GetStringLength(), CE_UTF8);
```

src/BqField.cpp
```diff
@@ -253,6 +270,30 @@ class BqField {
...
+      if (v.IsString()) {
+        size_t outlen = 0;
+        const unsigned char* buf = reinterpret_cast<const unsigned char*>(v.GetString());
+        unsigned char* out = base64_decode(buf, v.GetStringLength(), &outlen);
+        if (out == NULL) {
+          Rcpp::stop("Error parsing base64-encoded bytes");
+        }
+
+        Rcpp::RawVector raw(outlen);
+        memcpy(&(raw[0]), out, outlen);
```
DavisVaughan, line 291: `base64_decode()` allocates the memory for `out` with `malloc()`, so you will need to free it with `free(out)` after you have copied the memory to the `raw` vector with this `memcpy()` call.

DavisVaughan: Another issue. Say `base64_decode()` allocates an `out` object of length `outlen`. Then you try and allocate `raw`, but there isn't enough memory to allocate it. I _think_ that Rcpp would throw an error here, which would stop the program before you can `free(out)`, resulting in a memory leak. This would be rare, but would probably be good to try and handle, but I'm not sure what Rcpp does when it can't allocate because it runs out of memory.

DavisVaughan: I think that the cast from the `v.GetString()` result of `const char*` to `const unsigned char*` is probably fine. It will wrap negative char values around, but I think that is well defined and would probably do the right thing.

I imagine that `&(raw[0])` gives a `Rbyte*`, and you are copying memory to that from an `unsigned char*`. I think this is probably fine too, since `Rbyte` is typedef-ed to `unsigned char`

https://github.com/wch/r-source/blob/abb550c99b3927e5fc03d12f1a8e7593fddc04d2/src/include/Rinternals.h#L55

DavisVaughan: Okay, I asked and the C++ way to handle the malloc-ed `out` array is to wrap it in a small C++ class that has a destructor. That destructor would `free()` it, and it would automatically be called when the object went out of scope (i.e. at the end of the function or if an Rcpp error gets thrown).

An example I know of that does something like this is stringi. Here is a string buffer class that has a destructor that frees on destruction. Your constructor could just take the `unsigned char*` since the object is already allocated by `base64_decode()`.
https://github.com/gagolews/stringi/blob/9ef18d13b7523a9c50761d16bfd8007af90ed42a/src/stri_string8buf.h#L48-L78

src/BqField.cpp
paleolimbot: `RObject` matches the usage for `BQ_STRING`...I could update both?

src/BqField.cpp
paleolimbot: I appreciate the explanation here...I've switched this out for the processx parser (see below for a question I have about that...)

src/BqField.cpp
DavisVaughan: In that case, I think its fine as is. `RObject` is a catch-all and you aren't doing anything fancy with it, so it should be fine!

DavisVaughan:
Gabor approved so I trust his judgement. As far as I am aware, as long as all the `Rinternals.h` related things are kept in a separate C file (as is done here), then it should be fine

R/zzz.R
```diff
@@ -17,6 +17,11 @@
   toset <- !(names(defaults) %in% names(op))
   if (any(toset)) options(defaults[toset])
 
+  # Namespaces for objects created in Rcpp ---------------------------------
+  requireNamespace("hms", quietly = TRUE)
```
hadley, line 21: This is a bit odd; I think we'd usually use a dummy `@importFrom` if you just want to make sure that the namespace is attached. But this makes hms and wk packages required, which is probably not what we want. Can you explain a bit more what you're trying to do here?

hadley (commented):
One question to answer, and can you please add a bullet to the NEWS?

R/zzz.R
paleolimbot: If you remove these, the S3 print methods don't get loaded, so the tibble formatting can look off (format/pillar methods) or at worst fail (because `vec_is()` might return `FALSE` if vctrs methods aren't defined, or this was the case at the time that I wrote the PR). This is a way to keep these packages in `Suggests` and it's more explicit than an `@importFrom`, even if the package is in `Imports`. I've added a comment here to clarify the reasoning, but I have no strong feelings and could use a dummy `@importFrom` if you think this is better.

I followed the hms example for wk and put it in `Suggests`, but if these packages aren't installed, the methods won't be loaded and users might get cryptic-looking output and/or hard-to-diagnose errors (so maybe both hms and wk should be in Imports?). I'm trying to slim down the wk package but it does take a hot second to compile (and depends on Rcpp, if that matters).

hadley: I don't see how this lets you keep them in suggests; won't it error if on load if the packages aren't installed?

hadley: I'm not sure what the right way to resolve this is — maybe `bq_parse_files()` should scan the column classes and call `requireNamespace()` if needed? (Maybe displaying a warning if the package isn't installed?)

hadley:
Gah, sorry, of course you're right. I was just confused by how it prints:

```R
requireNamespace("sdfasdfdasddsfs")
#> Loading required namespace: sdfasdfdasddsfs
#> Failed with error:  ‘there is no package called ‘sdfasdfdasddsfs’’
```

But that's not actually an error 🙄

hadley:
But I like this approach better since you get a nice error. Thanks for working on this!

PR #398, opened 2020-08-04
Update the bq_perform_query to allow user to pass in time partitioning and clustering

SeagleLiu:
Time partitioning an clustering are important properties of bq table to reduce cost and improve performance. These two options will allow users to specify time partition and clustering when the new table is created from the query results.


SeagleLiu:
Fixed the roxygen issue.. Now all checks should have passed except for the mac dev one.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: bigrquery
 Title: An Interface to Google's 'BigQuery' 'API'
-Version: 1.3.1.9000
+Version: 1.3.1.90001152
```
hadley, line 3: Can you please leave this as is?

R/bq-perform.R
```diff
@@ -244,12 +244,16 @@ bq_perform_load <- function(x,
 #'   "INTERACTIVE" and "BATCH". Batch queries do not start immediately,
 #'   but are not rate-limited in the same way as interactive queries.
 #' @param default_dataset A [bq_dataset] used to automatically qualify table names.
+#' @param time_partitioning A list with following elements: type (string), filed (string), requirePartitionFilter (boolean), the later of each is converted in the as_json step.
```
hadley, line 247: Can you please provide a link to the BigQuery docs for these parameters?

R/bq-perform.R
```diff
@@ -244,12 +244,16 @@ bq_perform_load <- function(x,
...
+#' @param clustering A list of one element named fields, which is a vector of strings.
```
hadley, line 248: Why can't you supply the character vector directly?

hadley (commented):
Can you please also add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

byapparov:
@SeagleLiu, 

Just note that you can already create an empty table with partition using `bigrquery::bq_table_create()`:

```r
clustering <- list("field1")
day.partitioning <- list(type = "DAY")
bigrquery::bq_table_create(
      tbl,
      fields = read_json(schema.file),
      timePartitioning = day.partitioning,
      clustering = clustering
)
```

SeagleLiu:
@byapparov Thanks for pointing this out. This PR is meant to allow the timePartitioning to be specified when the table is created from queries. In our use cases, we are joining/filter some big tables to create new ones. The original bq_perform_query did not pass on these two parameters.

hadley:
If you're still interested in this, would you mind filing an issue? I think the current implementation is a bit too minimal, and I'm worried that it'll make it too easy for the user to shot themselves in the foot.

PR #400, opened 2020-08-17
Force BQ page query params to not use scientific notation

### LINKED ISSUE #395: intermittent error: Invalid value at 'start_index' (TYPE_UINT64), "1e+05" [invalid]
A report that I've been running daily for months started failing in the last day or two with an intermittent error. I can usually re-run and get it to work, but it's puzzling. It appears to be an error pulling the data back.

In `bq_check_response`, `result$content` has a 400 error code and the error `Error: Invalid value at 'start_index' (TYPE_UINT64), "1e+05" [invalid] `. `start_index` isn't in my code.

Happy to help debug this, but not sure where to start (or how to generate a reprex). 

bigrquery is version 1.3.1; R version 3.6.1; x86_64-apple-darwin15.6.0

gjuggler:
Fixes #395.

I work on the All of Us Researcher Workbench project (https://github.com/all-of-us/workbench) where we guide users to fetch data from BigQuery via the bigrquery package. The issue above has been affecting our users, so we wanted to take a swing at a fix.

**Rationale**
This is a fix with more limited scope than the `httpr` alternative (https://github.com/r-lib/httr/pull/670). We simply force the page params 

**Testing**

- devtools::check() runs successfully.
- I confirmed that the added regression test fails if the code change is commented out.
- Open to suggestions for an alternative test strategy. This approach (using opts to force aggressive scientific formatting) is a bit more clever than I'd like, but it prevents us from needing to load ~100k rows to trigger this scenario.

**Notes**
An alternative to the approach in this PR would have been to use `as.integer(x)` instead of `format(x, scientific=FALSE)`. But that alternative relies on the implicit difference in R's formatting of integers vs. doubles, so I opted for the more explicit fix. I'd be curious to hear from reviewers if `as.integer` would be more robust or idiomatic.

### DIFF
--- a/tests/testthat/test-bq-download.R
+++ b/tests/testthat/test-bq-download.R
@@ -31,16 +31,14 @@ test_that("can retrieve zero rows", {
 test_that("can specify large integers in page params", {
   skip_if_no_auth()
 
-  old_opts <- options()
   # Use scipen to nudge R to prefer scientific formatting for very small numbers
   # to allow this test to exercise issue #395 with small datasets.
-  options(scipen=-4)
+  old <- options(scipen=-4)
+  on.exit(options(old))
 
   tb <- as_bq_table("bigquery-public-data.moon_phases.moon_phases")
   df <- bq_table_download(tb, max_results = 100, page_size = 20)
   expect_equal(nrow(df), 100)
-
-  options(old_opts)
 })
 
 # bq_table_info -----------------------------------------------------------
--- a/tests/testthat/test-bq-download.R
+++ b/tests/testthat/test-bq-download.R
@@ -28,6 +28,20 @@ test_that("can retrieve zero rows", {
   expect_named(df, c("phase", "phase_emoji", "peak_datetime"))
 })
 
+test_that("can specify large integers in page params", {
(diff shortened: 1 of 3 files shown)


tests/testthat/test-bq-download.R
```diff
@@ -28,6 +28,20 @@ test_that("can retrieve zero rows", {
   expect_named(df, c("phase", "phase_emoji", "peak_datetime"))
 })
 
+test_that("can specify large integers in page params", {
+  skip_if_no_auth()
+
+  old_opts <- options()
+  # Use scipen to nudge R to prefer scientific formatting for very small numbers
+  # to allow this test to exercise issue #395 with small datasets.
+  options(scipen=-4)
```
hadley, line 37: A slightly safer pattern is to do:

```R
old <- options(scipen = -4)
on.exit(options(old))
```

As that puts the clean-up next to the "mess"-up and it's run even if the test terminates early due to an error

gjuggler: Done. Thanks!

hadley:
Thank you!

Issue #316, opened 2019-06-15
Way to see cost of query before execution a lá web interface?


adhi-r:
On the BigQuery web console, you can write a query and learn how much it would cost without actually running it. Like so:
![image](https://user-images.githubusercontent.com/13878159/59554348-41cf0300-8f67-11e9-8ff7-7a491ac5333e.png)

Is there a way to take a query and get this information without actually performing the query from the R console?

[This seems to be the most relevant API doc](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/get), and unfortunately it looks like you can only get `location` information about a given job.

lhagen-isb:
I'm also interested in being able to return a "dry run" of the processing information through R. There is a section on the Google BigQuery Docs about being able to view the amount of data that will be queried in `['BigQuery best practices: Controlling costs'](https://cloud.google.com/bigquery/docs/best-practices-costs#bigquery_query_dry_run-api)`. It would be helpful to be able to use it in conjunction with dplyr.

danielkupka:
I would also be interested in this feature.

j450h1:
This is possible through the BQ CLI as per below, so I don't see why it couldn't be implemented in this package (if you are comfortable with shell commands you could use this method until then - simply replace the section in double quotes with your query):

```bash
bq query --dry_run --use_legacy_sql=false "SELECT 'this is a' AS test"
> Query successfully validated. Assuming the tables are not modified, running this query will process 0 bytes of data.          
```

viniciuszendron:
Any news on this subject? 

In documentation, it is stated that additional arguments (...) are passed on to `bq_perform_query()`, which passes to the API (converting snake_case arguments to camelCase, if needed).  However, putting `dry_run` or `dryRun` as an argument at the end of the function do not work as expected, returning the result of a regular query instead.

I think issue #205 partially resolved this problem with the additional arguments, but it was closed and not merged.

hadley:
The problem is that if supply `dry_run = TRUE` it gets added to the top-level of the JSON, but it actually needs to be included in the `configuration` (https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfiguration).

This means that it will need to be deliberately plumbed into the `bq_perform` functions.

hadley:
And it returns completely different output, so it really needs to be a different function. Something like this, maybe:

```R
bq_perform_query_dry_run <- function(
                             query, billing,
                             ...,
[...]

PR #402, opened 2020-09-23
Add bq_perform_query_dry_run

Ka2wei:
Add bq_perform_query_dry_run as suggested in issue #316 after changing totalBytesBilled to totalBytesProcessed (since Billed is always 0). Output now corresponds to output in BQ Console (although they show GiB and MiB in the console rather than the GB and MB shown by pretty_bytes)


hadley:
Thanks!

adhi-r:
So cool, thanks for doing this y'all!

PR #416, opened 2020-11-04
Update `grepl` SQL translation to match REGEXP_CONTAINS signature

ahmohamed:
`grepl` function in base R is incorrectly translated to `REGEXP_CONTAINS` by direct replacement. However, both have reverse arguments order: 

 - `grepl(pattern, x)`
 - `REGEXP_CONTAINS(value, regexp)`

This PR fixes the order of arguments for `REGEXP_CONTAINS`.


hadley:
Thanks for working on this, and sorry for taking so long to merge it!

PR #420, opened 2020-11-11
Don't init .auth object until .onLoad

jcheng5:
Creating the .auth object directly at the top-level means the code
from gargle is snapshotted at build time.

This is needed for r-lib/gargle#157, but is a good change even without that PR.


hadley:
Oops, thanks!

PR #425, opened 2020-12-02
Add BigQuery Storage API capabilities

meztez:
I'm leaving this here.

I could not get it to cooperate on macOS and I do not own a mac so there is that.

For it to be truly be integrated, it would require a better conditional configure script so people can build the package without bigquery storage support and use somekind of internal function after that to install dependencies for support.

Maybe next year.


hadley:
Are you still interested in this? If so, it might be better to start with it in a separate package that bigrquery could then suggest.

meztez:
We have been using a separate package for quite a while (https://github.com/meztez/bigrquerystorage). It is quite fine for large scale operation like ours.

Making the compile tool chain work on windows/macOS is quite unpleasant to maintain, so I'm not investing more time to make it publishable to CRAN. Sorry.

hadley:
@meztez would you be interested in getting on to CRAN if we provided help on the windows/mac side?

meztez:
@hadley sure, I'm not going to turn down help. You can ask Neil how hard it was to get arrow to CRAN :), and arrow is a much more useful package. abseil and grpc are not fun to play with, sometimes you have to patch the packages on msys. Someone else's patch might break your build chain. But I'm willing to go along with it just for the experience I might get out of it.

PR #307, opened 2019-05-09
Rstudio ui connection

byapparov:
Closes #298

PR #431, opened 2021-02-02
Updated RStudio connections pane integration

### LINKED ISSUE #298: Display bq connection in Rstudio
RStudio 1.1< allows to display sets from the Sql DDBB's connections.

This works with DBI, for example the driver odbc::odbc. 
But not with bigrquery i.e. sets from bigquery don't display in the UI.

More info on the release, [here](https://support.rstudio.com/hc/en-us/articles/115010915687) and [here](https://rstudio.github.io/rstudio-extensions/rstudio-connections.html)

meztez:
Based on `odbc` package connections pane integration and #307, this will enable connections pane integration when using

```r
con <- bigrquery::dbConnect(bigquery(), "{project}")
```

Further optional improvement
* connection actions
* organisation, project, dataset icons
* distinguish between table type

Fixes #298


hadley:
Any suggestions on how I should test this?

meztez:
Let me rewrite some of this PR from the master branch. Then it's a matter of using 

```r
con <- bigrquery::dbConnect(bigquery(), "{project}")
```

within RStudio/Workbench.

hadley:
Thanks! I'll have a play around with it. What were you thinking of for organisation? I would've thought project would be a more natural top level category.

hadley:
I've made a few changes:

* I've removed the organisation level of the hierarchy since I don't see what it corresponds to (happy to add back when you explain)
* I use the project as the host; I think that's the right philosophy to keep connections isolated.
* I use the connection project to filter the available datasets
* I explored the idea of also filtering based on the project dataset, but decided that's ultimately a default dataset, and doesn't make sense to filter based on it (also I think if you did that you'd need to incorporate into the host)
* I've made general tweaks to get into tidyverse style

Thanks for all your work on this. It made it really easy for me to dig in!

meztez:
Damn I should have guessed you were also working on this.

I made similar changes this afternoon. Dropped organisation.

I added the type necessary to display view vs table. Adding a type to bq_table list structure, plus passing down dot arguments in as_bq_table method. etc...

I still have to add / support for the default icon in the pane. I'm trying to avoid having to put them in the package since rstudio ships with defaults one for database and schema (project and dataset).

Will get back to it when my kid is off to bed.

meztez:
> Thanks! I'll have a play around with it. What were you thinking of for organisation? I would've thought project would be a more natural top level category.

Since it was in was 2021, I'm inclined to respond something along the line of : `segfault : memory not mapped`.

hadley:
I can likely find out how to access the default catalog and schema icons if you think that those names might confuse folks who are used to the google terminology.  I'm also not sure if BigQuery now also can have schema nested inside dataset, or there's just a handful of special schema's like `INFORMATION_SCHEMA`.

meztez:
@hadley It's picked up here. Either an icon path is provided from the object type list or it uses a default one.

hadley:
Right, but it's picked up because you're using catalog and schema, which aren't BigQuery terms. Do you think that will be confusing to users?

meztez:
Yes, I did not like it either. I went ahead and reused the console UI icons.

image
image

Material icons
https://developers.google.com/fonts/docs/material_icons Licence Apache v2

hadley:
Looks great, thanks for working on this!

PR #440, opened 2021-04-06
collect now passes con bigint parameter to bq_table_download

### LINKED ISSUE #439: integer64 coercion and overflow with collect()
Moving [a dbplyr issue](https://github.com/tidyverse/dbplyr/issues/610) here with simplified reprex.

## The issue
I'm attempting to read a BigQuery table with an integer64 type column. I create a `DBI::dbConnect()` object using the `bigrquery::bigquery()` BigQuery DBI driver with `bigint` parameter set.  

We see divergent behavior when using `DBI::dbGetQuery` versus using `collect()`:
* When using `DBI::dbGetQuery` the resulting tibble returns the `bit64::integer64` type column as expected
* When using `collect()` the column is coerced to integer type, resulting in integer overflow and NAs

## Reprex
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(bigrquery)

# BigQuery authentication and project-specific names
bq_auth(path = ---)
project_name <- bq_projects()
dataset_name < ----
table_name <- ---
long_table_name <- paste0(dataset_name, '.', table_name)

# Both cases below use the same connection object
con <- DBI::dbConnect(
  bigrquery::bigquery(),
  project = project_name,
  dataset = dataset_name,
  billing = project_name,
  bigint = "integer64"
)

# Case one: DBI::dbGetQuery correctly returns tibble with integer64 column
output_tibble <- DBI::dbGetQuery(con, paste0('select * from ', long_table_name)) %>%
  glimpse()
#> Rows: 32
#> Columns: 1
#> $ int64_identifier <int64> 9223372036854775800, 9223372036854775800, 922337203…

# Case two: dbplyr attempts to coerce to integer, overflow results in NAs
tbl(con, long_table_name ) %>%
  select(int64_identifier) %>%
  collect() %>%
  glimpse()
#> Warning in as.integer.integer64(x): NAs produced by integer overflow
#> Rows: 32
#> Columns: 1
#> $ int64_identifier <int> NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, N…
```

<sup>Created on 2021-04-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

## Notes

As per @hadley in the `tidyverse/dbplyr` issue thread:
> (It looks like the root cause is that in collect.tbl_BigQueryConnection bq_table_download() doesn't get the value of the bigint field from x$src$con, so a PR should be straightforward if you want to give it a go).

I will work on a PR for this now. Thanks!

zoews:
Fixes #439 - dplyr snytax with `collect()` now returns bit64::integer64 values as expected when reading BigQuery table with int64 column:

```{r}
...

tbl(con, long_table_name ) %>%
  select(int64_identifier) %>%
  collect() %>%
  glimpse()
#> Rows: 32
#> Columns: 1
#> $ int64_identifier <int64> 9223372036854775800, 9223372036854775800, 922337203…
```
PR passes `devtools::check()`

### DIFF
--- a/R/dplyr.R
+++ b/R/dplyr.R
@@ -124,7 +124,7 @@ collect.tbl_BigQueryConnection <- function(x, ...,
   }
 
   quiet <- if (n < 100) TRUE else x$src$con@quiet
-  bigint <- if (!is.null(x$src$con@bigint)) x$src$con@bigint else "integer"
+  bigint <- x$src$con@bigint %||% "integer"
   out <- bq_table_download(tb,
     max_results = n,
     page_size = page_size,
--- a/R/dplyr.R
+++ b/R/dplyr.R
@@ -124,11 +124,13 @@ collect.tbl_BigQueryConnection <- function(x, ...,
   }
 
   quiet <- if (n < 100) TRUE else x$src$con@quiet
+  bigint <- if (!is.null(x$src$con@bigint)) x$src$con@bigint else "integer"
   out <- bq_table_download(tb,
     max_results = n,
     page_size = page_size,
     quiet = quiet,
-    max_connections = max_connections
+    max_connections = max_connections,
+    bigint = bigint
   )
   dplyr::grouped_df(out, intersect(dbplyr::op_grps(x), names(out)))
 }
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 * Add `billing` slot to `BigQueryResult`.
 
-* When using dplyr syntax with BigQuery table, `collect()` now utilizes `bigint` parameter in `DBI::dbConnect()` object. Set `bigint =  integer64` in connection object to avoid int64 coercision to integer and resulting overflow issues. (@zoews, #439)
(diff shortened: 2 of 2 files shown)


R/dplyr.R
```diff
@@ -124,11 +124,13 @@ collect.tbl_BigQueryConnection <- function(x, ...,
   }
 
   quiet <- if (n < 100) TRUE else x$src$con@quiet
+  bigint <- if (!is.null(x$src$con@bigint)) x$src$con@bigint else "integer"
```
hadley, line 127:
```suggestion
  bigint <- x$src$con@bigint %||% "integer"
```

`%||%` is a handy helper from rlang

hadley (approved):
Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

zoews:
Great, thank you @hadley! All ready to go. 

I appreciate your guidance on this (and feedback on the reprex process earlier) -- will definitely draw on this all for future issues/PRs :)

hadley:
Thanks!

PR #452, opened 2021-06-30
WIP switch to proper pagination

hadley:
For @jennybc's eyes only


jennybc:
FYI I rebased this on `main` and am studying it as part of my orientation. I know it does not pass tests (nor did it before).

PR #454, opened 2021-07-22
Re-build with current Rcpp

jennybc:
This appears with Rcpp 1.0.7.

Mostly doing this to keep it out of my PR re: table download pagination.


hadley (approved):

hadley:
This reminds me that we should switch to cpp11 at some point.

PR #456, opened 2021-07-26
Download data in chunks that are more likely to be complete

### LINKED ISSUE #412: bq_table_download not uploading a large dataset into R? 5M+ rows
I am querying a table in Google BigQuery in RStudio. The dataset is 5M+ rows. I don't receive an error when I query/upload the data, in fact it mentions that all the rows uploaded as a data frame/

Code & Ouput: 

```
sql_string = "SELECT * FROM table"
df_base <- bq_project_query(billing, sql_string)
df_base <- bq_table_download(df_base, max_results = Inf, page_size = 1000000)
Downloading 5,219,696 rows in 6 pages.
```

When I View(df_base)  the 5,219,696 dataset shows but at row 89,086 all the data in my cells are blank

jennybc:
Fixes #412

The original bug is that some requests for `n` rows contain fewer than `n` rows, because the response gets truncated based on its size instead of its number of rows. This resulted in a returned data frame that smells OK: it has the right shape and first few hundreds or thousands of rows look fine, but they are then followed by empty rows (as many as necessary to deliver something with the user's requested number of rows).

High-level conclusions:

* We want to remain in charge of breaking the work into chunks, because then we can do downloads in parallel, which is totally worth it here. Letting Google paginate, then creeping through `pageToken`s is slow and also makes parsing + combining much more fiddly.
* We can't truly eliminate the possibility of server-side pagination, but now we actively check for it and throw an error if we've failed to retrieve all the requested rows.
* In the future, instead of engaging in more gymnastics around the awkward pagination behaviour of `bq_table_download()`, it's probably time to use the BigQuery Storage API.

Key links:

* API reference docs for `tabledata.list`: <https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/list>
* Guide on "Paging through table data": <https://cloud.google.com/bigquery/docs/paging-results>
* Related Python workflow: <https://cloud.google.com/bigquery/docs/bigquery-storage-python-pandas>
* API docs for BigQuery Storage: <https://cloud.google.com/bigquery/docs/reference/storage>


R/bq-download.R
```diff
@@ -52,42 +52,133 @@
...
+    }
+
+    # break rest of work into natural chunks ----
+    chunk_size <- trunc(0.75 * n_got)
+    message(glue("First chunk has {n_got} rows."))
+
+    chunk_plan <- bq_download_plan(
+      n_max,
+      chunk_size = chunk_size,
+      # TODO: start where we left off?
```
jennybc, line 144: Currently I just start over from the beginning, but I certainly could use the data already in hand. This  will be a minor speed/efficiency gain.

R/bq-download.R
```diff
@@ -52,42 +52,133 @@
...
+      quiet = quiet
+    )
+
     if (!bq_quiet(quiet)) {
       message(glue_data(
-        page_info,
-        "Downloading {big_mark(n_rows)} rows in {n_pages} pages."
+        chunk_plan,
+        "Downloading {big_mark(n_max)} rows in {n_chunks} chunks \\
+         of (at most) {chunk_size[1]} rows."
```
jennybc, line 157: My messages are worded for me, at this point.

R/bq-download.R
```diff
@@ -52,42 +52,133 @@
...
+      )
+    }
+    curl::multi_run(pool = pool)
+    withr::defer(file.remove(chunk_plan$dat$path))
+
+    table_data <- bq_parse_files(
+      schema_path,
+      chunk_plan$dat$path,
+      n = chunk_plan$n_max,
+      quiet = bq_quiet(quiet)
```
jennybc, line 180: In the next piece of work, we have to alter this parsing phase so that we actually determine how many rows were retrieved for each chunk. When that is fewer than requested, we need to create new chunks that cover all these holes and download the remaining data.

The presence of `pageToken` is useless to us in terms of learning how much data is left. It will _always_ be there, until you have literally read the whole table.

R/bq-download.R
```diff
@@ -117,91 +208,56 @@ rapply_int64 <- function(x, f) {
...
-}
-
-bq_download_pages <- function(x, page_info, max_connections = 6L, quiet = NA) {
-  x <- as_bq_table(x)
-  assert_that(is.list(page_info))
-
-  n_pages <- page_info$n_pages
-  if (n_pages == 0) {
-    return(character())
+bq_download_plan <- function(n_max,
```
jennybc, line 211: `bq_download_plan()` is morally the successor to `bq_download_page_info()`. Differences:

* Vocabulary stresses _chunks_ (under our control) versus _pages_ (under Google's control).
* Packages the info needed for chunk-by-chunk work into a tibble vs. lists-of-equal-length.
* Stores chunk beginning (destined for `startIndex` query param) and chunk rows (destined for `maxResults` query param). No more `page_end`.
* Includes the temporary file paths.

R/bq-download.R
```diff
@@ -117,91 +208,56 @@ rapply_int64 <- function(x, f) {
...
-
-  list(
-    n_rows = max_results,
-    n_pages = n_pages,
-    begin = page_begin,
-    end = page_end
-  )
-}
-
-bq_download_pages <- function(x, page_info, max_connections = 6L, quiet = NA) {
```
jennybc, line 138: This got inlined. Mostly so I could store `pageToken`s from the callback function. Now that I no longer think that has value, I could probably abstract this out again. Depends on how we think we'll structure the second piece of work, which might re-introduce this need.

R/bq-download.R
```diff
@@ -117,91 +208,56 @@ rapply_int64 <- function(x, f) {
...
+  list(
+    n_max = n_max,
+    chunk_size = chunk_size,
+    n_chunks = n_chunks,
+    dat = dat
+  )
 }
 
-bq_download_page_handle <- function(x, begin = 0L, end = begin + 1e4) {
+bq_download_chunk_handle <- function(x, begin = 0L, max_results = 1e4) {
```
jennybc, line 246: I think the original logic (`end = begin + 1e4`) is wrong, presumably caused by the misleading documentation for `maxResults`.

R/bq-download.R
```diff
@@ -219,6 +275,17 @@ bq_download_page_handle <- function(x, begin = 0L, end = begin + 1e4) {
...
+  fragment <- rawToChar(readBin(result$content, n = 300, what = "raw"))
+  tmp <- strsplit(fragment, split = "\n")[[1]]
+  tmp <- grep("pageToken", tmp, value = TRUE)
+  if (length(tmp) == 0) {
+    return("")
+  }
+  tmp <- strsplit(tmp, split = ":")[[1]][[2]]
+  sub('.*["](.*)["].*', "\\1", tmp)
+}
+
```
jennybc, line 288: I no longer think we can really make use of these. But I'll leave this here for the moment to show how I excavated the tokens without parsing the whole file.

tests/testthat/test-bq-download.R
```diff
@@ -42,6 +41,7 @@ test_that("can specify large integers in page params", {
 # bq_table_info -----------------------------------------------------------
 
 test_that("max_results + start_index affects end values", {
+  skip("function no longer exists")
```
jennybc, line 44: I'll be back to soon to write some tests for the function that replaced this.

R/bq-download.R
```diff
@@ -53,41 +53,132 @@
 bq_table_download <-
   function(x,
            max_results = Inf,
```
jennybc, line 55: Ideally, we would deprecate this argument and replace it with `n_max`. But I would understand if this is too disruptive to consider.

R/bq-download.R
```diff
@@ -53,41 +53,132 @@
...
+
+        con <- file(chunk_plan$dat$path[i], open = "wb")
+        withr::defer(close(con))
+        writeBin(result$content, con)
+      }
+    }
+
+    # get first chunk ----
+    if (!bq_quiet(quiet)) {
+      message("Downloading first chunk of data.")
```
jennybc, line 109: It would be nice to report progress here.

R/bq-download.R
```diff
@@ -53,41 +53,132 @@
...
+    curl::multi_add(
+      handle,
+      done = bq_download_callback(1),
+      pool = pool
+    )
+    curl::multi_run(pool = pool)
+    chunk_data <- bq_parse_file(schema_path, chunk_plan$dat$path[1])
+    n_got <- nrow(chunk_data)
+
+    if (!nzchar(chunk_plan$dat$next_page[1]) || n_got >= n_max) {
```
jennybc, line 132: As I've said elsewhere, I've decided these `nextPage` tokens are of no use to us, so I think we can just go with the second condition.

R/bq-download.R
```diff
@@ -117,91 +208,56 @@ rapply_int64 <- function(x, f) {
...
+  chunk_begin <- start_index + (seq_len(n_chunks) - 1) * chunk_size
+  chunk_end <- pmin(chunk_begin + chunk_size, n_max)
+  chunk_rows <- chunk_end - chunk_begin
+  dat <- tibble::tibble(
+    chunk_begin,
+    chunk_rows,
+    path = sort(
+      tempfile(rep_len("bq-download-", length.out = n_chunks), fileext = ".json")
+    ),
+    next_page = rep_len("", length.out = n_chunks)
```
jennybc, line 235: I've lost interest in storing these now. Will remove once we've discussed this.

R/bq-download.R
```diff
@@ -53,41 +53,132 @@
 bq_table_download <-
   function(x,
            max_results = Inf,
-           page_size = 1e4,
+           page_size = NULL,
```
jennybc, line 56: I believe we should no longer have a default `page_size`, nor encourage users to set this (although that should remain allowed).

In this PR, we try to discover the "natural" page size, which I believe to vary considerably across datasets, in a hard-to-predict way. It's also not a number, but rather a distribution, influenced by the type of data in the table, sparsity, etc.

jennybc:
One of the reprexes in #412, where we now get the correct result:

``` r
library(bigrquery)
library(dplyr, warn.conflicts = FALSE)

con <-
  bigrquery::dbConnect(
    bigrquery::bigquery(),
    project = "bigquery-public-data"
  )

# reprex from GitHub ----
system.time(
  df <-
    tbl(con, "chicago_taxi_trips.taxi_trips") %>%
    head(2e6) %>%
    collect(page_size = 100000, bigint = "character")
)
#> ! Using an auto-discovered, cached token.
#>   To suppress this message, modify your code or options to clearly consent to
#>   the use of a cached token.
#>   See gargle's "Non-interactive auth" vignette for more details:
#>   <https://gargle.r-lib.org/articles/non-interactive-auth.html>
#> ℹ The bigrquery package is using a cached token for 'jenny@rstudio.com'.
#> Auto-refreshing stale OAuth token.
#> First chunk has 29221 rows.
#> Warning in as.integer.integer64(x): NAs produced by integer overflow

#> Warning in as.integer.integer64(x): NAs produced by integer overflow
#>    user  system elapsed 
#>  25.144  13.008  48.499

df %>%
  filter(unique_key == "") %>%
  nrow()
#> [1] 0
```

<sup>Created on 2021-07-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0.9000)</sup>

hadley (approved):
Overall strategy looks great. Didn't review the implementation.

R/bq-download.R
jennybc: Now de-inlined.

R/bq-download.R
jennybc: We've decided to detect a row shortage, in aggregate, within the parsing code and, if present, error.

R/bq-download.R
jennybc: Done.

DESCRIPTION
```diff
@@ -34,12 +34,13 @@ Imports:
...
     glue (>= 1.3.0),
     httr,
+    lifecycle,
     jsonlite,
     methods,
     prettyunits,
     progress,
     Rcpp,
-    rlang,
+    rlang (>= 0.4.9),
```
jennybc, line 43: I'm using error bullets.

R/dplyr.R
```diff
@@ -101,7 +101,7 @@ db_copy_to.BigQueryConnection <- function(con, table, values,
 
 # registered onLoad
 collect.tbl_BigQueryConnection <- function(x, ...,
-                                           page_size = 1e4,
+                                           page_size = NULL,
```
jennybc, line 104: I suspect there are other distant locations (other `collect()` methods?) where a fixed default `page_size` still lurks, whereas it would be better to pass `NULL` now.

jennybc:
I made this color-coded spreadsheet to analyze / document the situation:

https://docs.google.com/spreadsheets/d/12g0arjMfEuVRyJ-KBTGHUSCwAKIasI-IJ7bxMMXrSbY/edit#gid=0

I moved it out of the first comment so it's not part of the ultimate commit message, but maybe it's worth preserving.

PR #462, opened 2021-09-28
Add readr to Imports

AdeelK93:
I have my own package that I'm working on, and it happens to include a `R CMD CHECK` test that calls a `bigrquery` function.

This test has been failing because `bigrquery` has `readr` in the Suggests section of the DESCRIPTION rather than the Imports. Fortunately this is a quick fix.

Here is an example of where `readr` is used in `bigrquery`:
https://github.com/r-dbi/bigrquery/blob/main/R/bq-download.R#L340

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,7 +42,7 @@ Imports:
     Rcpp,
     rlang (>= 0.4.9),
     tibble,
-    readr
+    brio
 Suggests: 
     blob,
     covr,
@@ -50,6 +50,7 @@ Suggests:
     dbplyr (>= 1.3.0),
     dplyr (>= 0.7.0),
     hms,
+    readr,
     sodium,
     testthat (>= 2.1.0),
     withr,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -41,7 +41,8 @@ Imports:
     progress,
     Rcpp,
     rlang (>= 0.4.9),
-    tibble
+    tibble,
+    readr
 Suggests: 
     blob,
     covr,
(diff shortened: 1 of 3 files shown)


jennybc:
The function that you point out (`bq_parse_file()`) was originally commented as one of the "Helpers for testing". But we are now using it to parse the first chunk obtained in `bq_table_download()` (as of 1.4.0, thanks to my PR #456). Apparently I didn't realize that it made unconditional calls to a suggested package (readr).

@Hadley Do you have a sense how you want to handle this? I suspect importing readr is not the right solution. We're just using `readr::read_file()`, presumably for pragmatic reasons.

hadley:
Probably the easiest fix would be to use `brio::read_file()` instead; it should have similar performance characteristics and is a low-dependency package.

jennybc:
@AdeelK93 can you update this PR to:

* *Not* import readr.
* Import brio. Please note that we alphabetize such package-listing sections of DESCRIPTION.
* Replace calls to `readr::read_file()` with `brio::read_file()`

AdeelK93:
Thanks @jennybc @hadley.
I've made those changes, and left readr in Suggests

jennybc:
@Hadley Shall we do a patch release?

FWIW the only remaining uses of readr are via the deprecated `list_tabledata()` function:

https://github.com/r-dbi/bigrquery/blob/1a549b0b8fb7ab1a3584dd49ba8af64792e3f0a0/R/old-tabledata.R#L37-L39
https://github.com/r-dbi/bigrquery/blob/1a549b0b8fb7ab1a3584dd49ba8af64792e3f0a0/R/old-tabledata.R#L214-L222

Issue #468, opened 2021-11-09
Information Schema


muschellij2:
The issue is that you cannot send in `INFORMATION_SCHEMA.*` options in the current R API.  As per https://cloud.google.com/bigquery/docs/information-schema-tables#bq_4, you should be able to query `INFORMATION_SCHEMA.PARTITIONS`.   It seems as though the check to the `.` separators take issue with things like `INFORMATION_SCHEMA.PARTITIONS`.  You can pass the query either to `bq_dataset_query` or `dbGetQuery` so that you can get the result.  I believe this may be an edge case, but getting access to the schema may be important for some

``` r
library(bigrquery)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union


json_file = bigrquery::bq_auth(path = Sys.getenv("GCE_AUTH_FILE"))
dataset = "mydataset"
project = "myproject"

con <- DBI::dbConnect(
  bigrquery::bigquery(),
  project = project
)

table = "INFORMATION_SCHEMA.PARTITIONS"
bq_ds = bq_dataset(project, dataset)
bq_tab = bq_table(project, dataset, table = table)
bq_table_download(bq_tab)
#> Error: Illegal operation on INFORMATION_SCHEMA view: myproject:mydataset.INFORMATION_SCHEMA.PARTITIONS [invalid]
tab = tbl(con, paste0(dataset, ".", table))
#> Error: Invalid project ID 'mydataset'. Project IDs must contain 6-63 lowercase letters, digits, or dashes. Some project IDs also include domain name separated by a colon. IDs must start with a letter and may not end with a dash. [invalid]
tab = tbl(con, paste0(project, ".", dataset, ".", table))
#> Error: Table name, 'myproject.mydataset.INFORMATION_SCHEMA.PARTITIONS', must have 1-3 components.

query = paste0(
  "SELECT * FROM ", dataset, ".", table)

res = DBI::dbGetQuery(con, query)
dim(res)
#> [1] 2 9

result = bq_dataset_query(bq_ds, query = query)
result
#> <bq_table> myproject._afedd2ea8c5ca00d5c8dab2ccd9eb21581bd8cd3.anonev_7E6lIwQmlcn1uvJAYnzPOAzEqvBV3P0ieyz0Zzn6mCE
dim(bq_table_download(result))
[...]

PR #469, opened 2021-11-09
Added in meta-data helper extractors

muschellij2:
Added in helpers to extract metadata, related to #468.  Simple helper to extract modification time.


muschellij2:
I think the `mtime` accessor would be used frequently to see if a local data source needs to be refreshed (e.g. in targets)

hadley:
It feels a bit weird to add a bunch of simple wrappers around `bq_table_meta()`. Would you mind opening an issue to discuss the problem that you're trying to solve?

muschellij2:
Sure, main "issue" is that people have to read docs on how to convert times to POSIXct, but could be done in `bq_table_meta` itself:
```r
  as.POSIXct(as.double(meta$TIME_VARIABLE)/1000,
             origin = "1970-01-01", tz = "UTC")
```

PR #480, opened 2022-01-19
make bq_table_create() friendlier by using x as default fields.

taiyodayo:
This prevents unfriendly cast error messages from appearing, by defaulting to using the fields/coltypes from input dataframe x.

This SO thread inspired me to make this proposal.
https://stackoverflow.com/questions/53440116/bigrquery-forcefully-coerces-strings-to-integers-schema-is-a-string
( I myself took a while to be able to send data to BQ, like the original poster. )

This is my first PR. Please accept my apologies if anything is not appropriate!
Thank you for the wonderful package.


taiyodayo:
Sorry wrong file was uploaded. canceling PR

PR #489, opened 2022-03-21
adding support for location in bq_perform_query

ben-tanen:
Adding support to specify a query/job location, as outlined in [the BigQuery Jobs documentation](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#QueryRequest). Similar changes could likely be made elsewhere, but this is a start.


hadley:
Would you mind starting with an issue? I think there's a lot of additional parameters that people might want to add, and I want to make sure that we don't paint ourselves into a corner by only considering them one at a time. Apologies for taking so long to get this feedback to you 😞

PR #493, opened 2022-05-27
Add support for BIGNUMERIC data type

### LINKED ISSUE #435: Data Type: BIGNUMERIC support
There is a new decimal type `BIGNUMERIC` available for Big Query tables (see [data types docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types)). When we load a table with this type, we receive the error message:

```r
Running job 'example' [\]  1s
Complete
Billed: XX.XX MB
Downloading Y rows in Z pages.
Error in bq_parse_files(schema_path, page_paths, n = page_info$n_rows,  :                              
  Unknown type BIGNUMERIC
```

I think that we can just map BIGNUMERIC to float type (`Rcpp::DoubleVector`), but I'm not a master here. A similar request in the python SDK can help [python-bigquery/ issue 367](https://github.com/googleapis/python-bigquery/issues/367).

Thank you for the great work!

OssiLehtinen:
Implementing support for Google BigQuery's BIGNUMERIC data type, by replicating the handling of the FLOAT type.

Fixes #435

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 # bigrquery (development version)
 
-brio is new in Imports, replacing the use of the Suggested package readr, in `bq_table_download()` (@AdeelK93, #462).
+* brio is new in Imports, replacing the use of the Suggested package readr, in `bq_table_download()` (@AdeelK93, #462).
+
+* Add support for BIGNUMERIC data type (@OssiLehtinen, #435)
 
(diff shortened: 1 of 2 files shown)


hadley:
A bignumeric is 32 bytes, but a float is only 8 bytes, so I don't think this will work in general 😞.  I appreciate you pitching in!

Issue #496, opened 2022-06-10
Library should set X-Goog-User-Project header to quota project


prodonjs:
When the `billing` property is set for a `BigQueryConnection`, it does not appear to be used in the `X-Goog-User-Project` request header as is described at https://cloud.google.com/apis/docs/system-parameters#google_private_system_parameters to associate the request with the appropriate billing project.

The example below shows that despite the billing project ID being set to foo, the request gets associated with a different project ID since it is not explicitly specified in the request header.

---
```r
library(DBI)
> 
> Sys.setenv(BIGQUERY_TEST_PROJECT = "foo") 
> 
> bigrquery::bq_auth("user@gmail.com")  
> 
> con <- dbConnect(
+   bigrquery::bigquery(),
+   project = "foo",
+   dataset = "dataset",
+   billing = bigrquery::bq_test_project()
+ )
> 
> httr::with_verbose(qry <- dbSendQuery(con, "SELECT * FROM `foo.dataset.test` LIMIT 10"))
-> POST /bigquery/v2/projects/foo/jobs/?fields=jobReference HTTP/2
-> Host: www.googleapis.com
-> user-agent: bigrquery/1.4.0 (GPN:RStudio; ) gargle/1.2.0 httr/1.4.2
-> accept-encoding: deflate, gzip, br
-> accept: application/json, text/xml, application/xml, */*
-> content-type: application/json
-> authorization: Bearer ...
-> content-length: 289
-> 
>> {
>>   "configuration": {
>>     "query": {
>>       "query": "SELECT * FROM `foo.dataset.test` LIMIT 10",
>>       "useLegacySql": false,
>>       "priority": "INTERACTIVE",
>>       "defaultDataset": {
>>         "projectId": "foo",
>>         "datasetId": "dataset"
>>       }
>>     }
>>   },
>>   "quiet": null
>> }
 
<- HTTP/2 403 
<- vary: Origin
<- vary: X-Origin
<- vary: Referer
<- content-type: application/json; charset=UTF-8
<- content-encoding: gzip
<- date: Thu, 19 May 2022 11:49:24 GMT
<- server: ESF
<- cache-control: private
<- content-length: 536
<- x-xss-protection: 0
<- x-frame-options: SAMEORIGIN
<- x-content-type-options: nosniff
<- 
Error in `signal_reason()`:
[...]

PR #504, opened 2022-08-23
Make bigrquery work in projects without BQ access

htappen:
This is a fix for issue: https://github.com/r-dbi/bigrquery/issues/496

Basic idea is to open the application default credentials for the quota project ID (which is usually set by gcloud), and add it to the headers.

### DIFF
--- a/R/bq-auth.R
+++ b/R/bq-auth.R
@@ -3,6 +3,7 @@
 
 # Initialization happens in .onLoad
 .auth <- NULL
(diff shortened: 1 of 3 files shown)


hadley:
Would you minding giving me a few more details/pointers to docs as to what makes this work?

hadley:
I haven't heard anything in a month so I'm going to close this PR. But please ping me if you're still interested and we can figure it out down the road.

PR #505, opened 2022-09-01
dbplyr compat part 2

mgirlich:
There were 3 issues:

1. The warning about the old dbplyr interface. I simply wrapped this in `expect_warning()` as I won't migrate bigrquery to dbplyr edition 2 (as I'm not using bigquery at all...).
2. The expected errros in `bq-job-errors.txt`. One changed slightly due to rlang. The other does not error anymore. Probably bigquery does not throw an error anymore. I added a note that this should be updated but this should be someone more experienced with bigquery.
3. The last error seems to be random. `bq_test_dataset()` is successfully called  a couple of times before and I cannot reproduce it either.

### DIFF
--- a/tests/testthat/bq-job-errors.txt
+++ b/tests/testthat/bq-job-errors.txt
@@ -1,11 +1,7 @@
 > # One error
 > bq_dataset_query(ds, "SELECT 1 +", quiet = TRUE)
-Error: Job '' failed
+Error in bq_job_wait(job, quiet = quiet): Job '' failed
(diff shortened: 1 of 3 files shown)


mgirlich:
@jennybc Here is the 2nd PR which hopefully works 😄

jennybc:
Hmmm, it looks like the live API job can't access secretes when run from an external PR, even if I push. That's not a big surprise. So I'll just merge this, because it clearly looks like progress, and we'll see what CI says once it's in `main`. Thanks for continuing to work on it.

PR #507, opened 2022-09-01
Update api-table documentation

calebwpalmer:
Arguments in the `bq_table_create()` example in api-table documentation were incorrectly assigned. Updated example uses bq_table() for x and mtcars dataset for fields.


jennybc (commented):
I'm not the maintainer of bigrquery, but do pitch in from time to time. I have some high-level observations on this PR that might make it more likely to be merged. Or, perhaps, make a new, smaller PR more likely to be merged.

From the title I expected a small documentation change, but then the actual changes seem quite substantial.

Is there a version of your proposed fix that's much more narrow and doesn't change dependencies and add new imports/exports?

jennybc:
Since the PR doesn't link back to a specific issue, it would also benefit from a bit more description/justification.

https://tidyverse.github.io/code-review/author/submitting.html

PR #532, opened 2023-04-17
Handle the package-level doc in a more modern way

jennybc:
This came from a trial run of `use_tidy_dependencies()`, which I ultimately backed out of. But some of this seemed worth doing in the spirit of spring cleaning and making some updates to make bigrquery closer to how we'd do things now. I'll make some observations inline.


DESCRIPTION
```diff
@@ -35,7 +35,8 @@ Imports:
     progress,
     Rcpp,
     rlang (>= 0.4.9),
-    tibble
+    tibble,
+    withr
```
jennybc, line 39: We were already using withr unconditionally below `R/`, so now that works by design instead of via coincidental indirect hard dependencies.

NAMESPACE
```diff
@@ -174,8 +174,9 @@ exportMethods(dbSendQuery)
 exportMethods(dbWriteTable)
 exportMethods(show)
 import(DBI)
-import(assertthat)
+import(assertthat, except = has_name)
 import(methods)
+import(rlang, except = unbox)
```
jennybc, line 179: Avoiding some name collisions.

R/bigrquery-package.R
```diff
@@ -9,14 +9,23 @@
...
-#' @import assertthat
-#' @importFrom jsonlite unbox
-#' @importFrom Rcpp sourceCpp
-#' @importFrom tibble tibble
-#' @importFrom bit64 integer64
-#' @importFrom rlang %||%
-#' @importFrom glue glue glue_data glue_collapse
-#' @importFrom lifecycle deprecated
 #' @keywords internal
+#' @aliases bigrquery-package
```
jennybc, line 13: bigrquery is one of those "fun" packages where there's also a function named "bigrquery".

R/bq-download.R
```diff
@@ -147,7 +147,7 @@ bq_table_download <-
     }
 
     if (chunk_size_from_user && n_got < chunk_size) {
-      rlang::abort(c(
+      abort(c(
```
jennybc, line 150: Now that we import cli, many `abort()` calls could be expressed more nicely with `cli_abort()`, but I don't feel like doing that now.

hadley (approved):

PR #538, opened 2023-07-07
use client= argument in token_fetch()

MichaelChirico:
I'm not 100% sure this is correct, but it certainly looks like the right consistency change to follow up 8460e24445d5434907702ea866d79d4052893a84.

I can only tell that the `app=` argument is marked `deprecated()` in `credentials_user_oauth2()`:

https://github.com/r-lib/gargle/blob/202fcaacd1151e9b9881ea154193bb50bd476194/R/credentials_user_oauth2.R#L54

And seemingly ignored in all the other default `credentials*` functions:

https://github.com/r-lib/gargle/blob/202fcaacd1151e9b9881ea154193bb50bd476194/R/cred_funs.R#L142-L151

But I may be missing something.


jennybc:
Yeah all of the gargle-downstream packages eventually need a second and final adjustment for the "app" to "client" transition.

An example of the heart of such an update is here in googledrive, which I'm noting for the benefit of whoever finishes this off, whether be it @MichaelChirico or me or Hadley:

https://github.com/tidyverse/googledrive/commit/4ddfde6491cef4827fd91e832f8a2beeda340d0b

So the current change in this PR is good, but I'd say there are other changes that would also make sense to do at the same time.

MichaelChirico:
Thanks for the link! Had a go at finding other 'app' usages, I think that covers everything. I searched for `(^|[^\w.])app($|[^\w.])` in the whole repo and found these two R files, Rd references to the deprecated arguments, and a NEWS entry discussing deprecation.

hadley:
Thanks!

hadley:
Thanks!

PR #572, opened 2023-11-07
Use modern gargle secret management

hadley:
@jennybc I don't think this code is in quite the right place, because ideally I want it to auth automatically for R CMD check and pkgdown. Can you suggest where I should move this code?


hadley:
Based on run time (60s vs ~16s on main), I'm pretty confident that this is now enabled for pkgdown too.

PR #443, opened 2021-04-10
Parameter support for DBI dbGetQuery

byapparov:
Closes: #444

This would help to implement something similar to [BigQuery Magics](https://googleapis.dev/python/bigquery/latest/magics.html) in Rmarkdown https://github.com/yihui/knitr/issues/1867

PR #573, opened 2023-11-07
Add support for parameterised queries

### LINKED ISSUE #444: Parameterised query support for DBI interface
I need to pass parameters to the query via DBI interface to implement parameterised query integration in knitr . 

```r
DBI::dbGetQuery(con, "SELECT @value", params = (value = 1))
```

hadley:
Build off #443. Fixes #444


hadley:
@byapparov thanks for contributing this code and apologies for taking so long to merge it!

PR #574, opened 2023-11-08
Handle 503 responses in bq_job_wait()

### LINKED ISSUE #535: bq_job_wait signals errors when a jobs.get call return a 503
Sometimes, not always, it takes a couple seconds for BigQuery to register a job.

If you poll the jobs.get endpoint too fast, you might receive a 503 http status with a backendError, which is normal per https://cloud.google.com/bigquery/docs/error-messages.

The `bq_job_wait` method is supposed to loop until the job is done or fails.
Since the first call to the endpoint resulted in a 503, it ends up aborting.
It does not get to the while loop.

```
bq_job_wait > bq_job_status > bq_job_meta > bq_get > process_request > bq_check_response (503) > signal_reason > abort
```

I'm unsure about the best way to address this issue.

Google suggest waiting a few seconds before retrying to pool the jobs.get api.
I'm thinking of adding a single conditional 2 seconds sleep on 503 to the jobs.get API.
It feels kinda hacky.

Would you have an opinion on the best way to resolve this?

hadley:
Fixes #535

@meztez would you mind taking a look at this/trying it out in your setup? I don't have an easy way to test it so it would be good to have another pair of eyes on it.


meztez:
Error we get on our side when it happen :
```r
Error in `signal_reason()`:
! Error encountered during execution. Retrying may solve the problem. [backendError]
Run `rlang::last_trace()` to see where the error occurred.
> rlang::last_trace()
<error/bigrquery_backendError>
Error in `signal_reason()`:
! Error encountered during execution. Retrying may solve the problem. [backendError]
---
Backtrace:
     ▆
  1. └─extract.data::extract_policies(...)
  2.   ├─base::do.call(extract_, c(extract_args, extraw_fun = extraw::get_policies))
  3.   └─extract.data (local) `<fn>`(...)
  4.     ├─base::do.call(extraw_fun, extraw_args)
  5.     └─extraw (local) `<fn>`(...)
  6.       ├─extraw:::get_policies_exec(...)
  7.       └─extraw:::get_policies_exec.BigQueryConnection(...)
  8.         └─extraw::dispatch_query(con, query0, type = "exec", tablename = tablename)
  9.           ├─extraw:::dispatch_query_exec(...)
 10.           └─extraw:::dispatch_query_exec.BigQueryConnection(...)
 11.             ├─bigrquery::bq_job_wait(...)
 12.             │ └─bigrquery::as_bq_job(x)
 13.             └─bigrquery::bq_perform_query(...)
 14.               └─bigrquery:::bq_post(url, body = bq_body(body, ...), query = list(fields = "jobReference"))
 15.                 └─bigrquery:::process_request(req)
 16.                   └─bigrquery:::bq_check_response(status, type, content)
 17.                     └─bigrquery:::signal_reason(json$error$errors[[1L]]$reason, json$error$message)
```

meztez:
@hadley it is still going to fail here
https://github.com/r-dbi/bigrquery/blob/2733bef59014efa4bd246dec87335863085b5d63/R/bq-job.R#L85
Before entering the while loop.

See #578 for reworked loop.

hadley:
And I forgot that bigquery has it's own error handling mechanism. Thanks!

PR #585, opened 2023-11-13
Switch to cpp11

hadley:


src/BqField.cpp
```diff
@@ -267,23 +269,26 @@ class BqField {
...
-      out.attr("class") = Rcpp::CharacterVector::create("tbl_df", "tbl", "data.frame");
-      out.attr("row.names") = Rcpp::IntegerVector::create(NA_INTEGER, -n);
+      out.attr("class") = {"tbl_df", "tbl", "data.frame"};
+      out.attr("row.names") = {NA_INTEGER, -n};
 
-      if (n == 0)
+      if (n == 0) {
+        out.attr("names") = names;
         return out;
+      }
```
DavisVaughan, line 278: Just remove this early exit entirely?

hadley: Good idea.

PR #613, opened 2024-06-11
Fix bq_table_upload when source_format is undefined

apalacio9502:
Hi @hadley,

This pull request aims to address the error encountered in the live API. This error occurs because when source_format is not defined, a vector of length 2 is being passed to unbox. Therefore, the correction for this error involves updating the source_format variable when arg_match is performed.

Regards,


hadley:
Nice catch, thanks!

PR #620, opened 2024-10-01
bigrquery storage is now on CRAN

hadley:


hadley:
@meztez this is still failing on R 4.1. Did the install code not make it in the CRAN release?

meztez:
CRAN does not allow external download for installation. I've kept it on GitHub though.

hadley:
Right. I guess I thought that you'd still include the code because it'll never be run on CRAN, because CRAN doesn't do run test packages on older versions of R. I think otherwise the fix doesn't help much because most people will be building the package from the CRAN version, not the GitHub version.

meztez:
i will do a re release then, test this hypothesis out

hadley:
Thanks @meztez!

meztez:
Released 1.2.1

hadley:
@meztez looks good. Thanks for all your work on this! I know this feature is going to make a lot of people happy.

meztez:
What are you talking about? You are the general that made all of this happen. Thank YOU.

PR #630, opened 2025-03-17
Remove obsolete isFALSE mask

MichaelChirico:
The package depends on R>=4.0, and `isFALSE()` comes with R since 3.5.0.

Keeping two slightly-different copies brings no more benefit (except maybe edge case back-compatibility) but could lead to hard-to-debug issues arising from differences between the two implementations.


hadley (approved):
I'll merge when I'm next working on bigrquery and can fix the build issues.

hadley:
Thanks!

Issue #660, opened 2026-01-06
POSIXt: tzone and differences


r2evans:
When I upload `POSIXt` that has no timezone set, it is interpreted differently on the round-trip query.

```r
bqcon <- DBI::dbConnect(bigrquery::bigquery(), ...) # the connection works
ds <- bigrquery::bq_table(..., table = "r2test")
tbl <- tibble(bool=c(T,F,NA), int=c(0:1, NA), dbl=c(0, 1, NA), chr=c("Yay","Nay",NA),
              dat=c(rep(Sys.Date(),2),NA), psx=c(rep(Sys.time(), 2), NA)) |>
  mutate(psxdbl = as.numeric(psx))
bigrquery::bq_table_upload(ds, tbl)
tbl2 <- DBI::dbGetQuery(bqcon, "select * from r2test") 
# all tables are inserted _backwards_?
tbl2 <- select(tbl2, all_of(names(tbl)))
all.equal(tbl, tbl2, check.attributes = FALSE)
# [1] "Component “psx”: 'tzone' attributes are inconsistent ('' and 'UTC')" "Component “psx”: Mean absolute difference: 18000.06"                

tbl
# # A tibble: 3 × 7
#   bool    int   dbl chr   dat        psx                      psxdbl
#   <lgl> <int> <dbl> <chr> <date>     <dttm>                    <dbl>
# 1 TRUE      0     0 Yay   2026-01-06 2026-01-06 21:05:53 1767733553.
# 2 FALSE     1     1 Nay   2026-01-06 2026-01-06 21:05:53 1767733553.
# 3 NA       NA    NA <NA>  NA         NA                          NA 
tbl2
# # A tibble: 3 × 7
#   bool    int   dbl chr   dat        psx                      psxdbl
#   <lgl> <int> <int> <chr> <date>     <dttm>                    <dbl>
# 1 TRUE      0     0 Yay   2026-01-06 2026-01-06 16:05:53 1767733553.
# 2 FALSE     1     1 Nay   2026-01-06 2026-01-06 16:05:53 1767733553.
# 3 NA       NA    NA <NA>  NA         NA                          NA 
tbl$psx[1] - tbl2$psx[1]
# Time difference of 5.000018 hours
```

When I do the same thing but with `"tzone"` set on the data, it is different (and more correct, within 65 milliseconds).

```r
DBI::dbExecute(bqcon, "drop table r2test")
attr(tbl$psx, "tzone") <- "UTC"
[...]

Issue #320, opened 2019-06-20
Need some way to control `digits` used by jsonlite


philbrierley:
I am finding that when uploading a data.table to big query, all values seem to end up in big query but rounded to 5 decimal places. Below is the call I'm making. The same data uploaded by other libraries has no such issue. Any pointers appreciated.

bq_table_upload(x=myBQTable, values=BQ_data_table, create_disposition='CREATE_NEVER', write_disposition='WRITE_APPEND')

byapparov:
@philbrierley, do you posing repex with data to make it easier to investigate this. 

Also, an you clarify 
>The same data uploaded by other libraries has no such issue. Any pointers appreciated.

Does it mean that uploading `data.frame` through bigrquery does not have rounding, or uploading `data.table` to bigquery through other R package does not have rounding?

philbrierley:
hi @byapparov, thanks for the response.

1. it's hard to give a reproducible example as you wouldn't have access to my bigquery account.  just try uploading a data.table to bigquery  using 

bq_table_upload(x=myBQTable, values=BQ_data_table, create_disposition='CREATE_NEVER', write_disposition='WRITE_APPEND')

where BQ_data_table is the data table, making sure some of the values have lots of decimal place.

2. Just to clarify, when I say other libraries, what I mean is non-R based methods.

byapparov:
ok. I can reproduce this, with both `data.frame` and `data.table`.

jennybc:
Related to https://github.com/r-lib/httr/issues/153, https://github.com/r-lib/httr/issues/455, and https://github.com/r-lib/httr/issues/539.

Yes, some way needs to be chosen to expose `toJSON()`'s `digits` to the motivated bigrquery user.

philbrierley:
Thanks for finding the issue. As my data set has some small numbers then a digits parameter would be much welcomed.

aldomann:
Any updates regarding this issue?

ncriswell:
The `default` package allows for the ability to change the default argument of a function. I have tried `default(toJSON) <- list(digits = 10)` but that doesn't seem to work. `stream_out` only has a `digits` argument in the `...` which seems invisible to the `default` function. Just wanted to add this in case it spurs any thoughts in others.

chucheria:
Any updates regarging this issue?

PR #662, opened 2026-01-07
fix 320, 660, and 661 (round-trip values relating to digits and tzone)

r2evans:
A few facets, all related to round-trip handling of numbers and their precision.

Before this patch:

```r
bq_auth(path="my_token.json")
tbl <- data.frame(dbl=pi, psx=Sys.time())
ds <- bigrquery::bq_table(...)
bq_table_upload(ds, tbl); tbl2 <- bq_table_download(ds); bq_table_delete(ds)
tbl$psx - tbl2$psx
# Time difference of 5.000016 hours
```

The "5 hours" issue is due to missing the tzone on upload. By appending the timezone on uploaded `POSIXt` data, we get then

```r
bq_table_upload(ds, tbl); tbl3 <- bq_table_download(ds); bq_table_delete(ds)
tbl$psx - tbl3$psx
# Time difference of 0.0591259 secs
```

where 59 milliseconds can be an issue for those needing sub-second precision. (It crushes my work completely.)

Additionally,

```r
tbl$dbl - tbl2$dbl
# [1] 2.65359e-06
```

That's not as troublesome, but it has been discussed and requested that there be a mechanism to set the `digits=`/precision of uploaded data. After this patch:

```r
options(bigrquery.digits=22, bigrquery.digits.secs=6)
bq_table_upload(ds, tbl); tbl4 <- bq_table_download(ds); bq_table_delete(ds)
tbl$dbl - tbl4$dbl
# [1] 0
tbl$psx - tbl4$psx
# Time difference of 9.536743e-07 secs
```

### DIFF
--- a/R/bq-perform.R
+++ b/R/bq-perform.R
@@ -193,7 +193,8 @@ export_json <- function(values) {
     if (is.null(attr(x, "tzone")) || attr(x, "tzone") %in% c(NA, "")) {
       attr(x, "tzone") <- Sys.timezone()
     }
-    paste(format(x, paste0("%Y-%m-%d %H:%M:%OS", digsec)), attr(x, "tzone"))
+    ifelse(is.na(x), NA_character_,
+           paste(format(x, paste0("%Y-%m-%d %H:%M:%OS", digsec)), attr(x, "tzone")))
   })
 
   # Convert wk_wkt to text


r2evans:
@hadley @jennybc 

Is there a way to escalate discussing this or the related bugs #320 and #660? I have premium Posit Support (I'm the POC for Ford Racing's use of Connect). This bug has significant impact, there's no way we can write anything to BQ when our client's data requires millisecond resolution.

hadley:
Anything you can do to make it faster for us to review a PR is always appreciated 😄

PR #673, opened 2026-02-23
Adding job labels for FinOps

### LINKED ISSUE #652: Add support for labels in BigQuery jobs
Currently, BigQuery jobs created through this package do not support passing labels. Adding support for labels would make it easier to track and manage jobs across teams and environments.

One way it could be done is by adding the labels as options, and sending them in the body configuration:
```diff
defaults <- list(
  bigrquery.quiet = NA,
-  bigrquery.page.size = 1e4
+  bigrquery.page.size = 1e4,
+  bigrquery.labels = NULL
)
```
```diff
- body <- list(configuration = list(query = query))
+ labels <- getOption("bigrquery.labels")
+ body <- list(
+   configuration = list(
+     query = query,
+     labels = labels
+   )
+ )
```

With a labels check and documentation and testing.

JulianUmbhau:
Added labels to job creation calls for FinOps purposes
Labels can be set globally via:
options(bigrquery.labels = list(env = "prod", team = "analytics"))
Labels are then automatically attached to all BigQuery job requests (query, load, extract, copy).
- Added check_labels() to validate labels against BigQuery's key/value constraints before sending
- Added bigrquery.labels default option (NULL) in .onLoad()
- Added tests for check_labels()
- Fixed pre-existing inconsistency: auto_unbox = TRUE was already present in bq_post/bq_patch but was missing from the toJSON calls in bq_perform_upload and bq_parse_single

Fixes #652

### DIFF
--- a/R/bq-parse.R
+++ b/R/bq-parse.R
@@ -1,6 +1,6 @@
 bq_parse_single <- function(value, type, ...) {
   field <- bq_field("", type, ...)
-  field_j <- jsonlite::toJSON(as_json(field), auto_unbox = TRUE)
+  field_j <- jsonlite::toJSON(as_json(field))
   value_j <- jsonlite::toJSON(value, auto_unbox = TRUE)
 
   bq_field_init(field_j, value_j)
--- a/R/bq-parse.R
+++ b/R/bq-parse.R
@@ -1,6 +1,6 @@
 bq_parse_single <- function(value, type, ...) {
   field <- bq_field("", type, ...)
-  field_j <- jsonlite::toJSON(as_json(field))
+  field_j <- jsonlite::toJSON(as_json(field), auto_unbox = TRUE)
   value_j <- jsonlite::toJSON(value, auto_unbox = TRUE)
 
   bq_field_init(field_j, value_j)
--- a/R/bq-perform.R
+++ b/R/bq-perform.R
@@ -168,7 +168,7 @@ bq_perform_upload <- function(
   metadata <- bq_body(metadata, ...)
   metadata <- list(
     "type" = "application/json; charset=UTF-8",
-    "content" = jsonlite::toJSON(metadata, auto_unbox = TRUE, pretty = TRUE)
+    "content" = jsonlite::toJSON(metadata, auto_unbox = TRUE, pretty = TRUE, digits = json_digits)
   )
 
   if (source_format == "NEWLINE_DELIMITED_JSON") {
--- a/R/bq-perform.R
+++ b/R/bq-perform.R
@@ -72,6 +72,8 @@ bq_perform_extract <- function(
   check_bool(print_header)
   check_string(billing)
 
+  labels <- check_labels(getOption("bigrquery.labels"))
+
   url <- bq_path(billing, jobs = "")
   body <- list(
     configuration = list(
@@ -81,7 +83,8 @@ bq_perform_extract <- function(
         destinationFormat = unbox(destination_format),
         compression = unbox(compression),
         printHeader = unbox(print_header)
-      )
+      ),
+      labels = labels
     )
   )
 
@@ -140,6 +143,8 @@ bq_perform_upload <- function(
   check_string(billing)
   json_digits <- check_digits(json_digits)
 
+  labels <- check_labels(getOption("bigrquery.labels"))
+
   load <- list(
     sourceFormat = unbox(source_format),
     destinationTable = tableReference(x),
@@ -154,11 +159,16 @@ bq_perform_upload <- function(
     load$autodetect <- unbox(TRUE)
   }
 
-  metadata <- list(configuration = list(load = load))
+  metadata <- list(
+    configuration = list(
+      load = load,
+      labels = labels
+    )
+  )
   metadata <- bq_body(metadata, ...)
   metadata <- list(
     "type" = "application/json; charset=UTF-8",
-    "content" = jsonlite::toJSON(metadata, pretty = TRUE, digits = json_digits)
+    "content" = jsonlite::toJSON(metadata, auto_unbox = TRUE, pretty = TRUE)
   )
 
   if (source_format == "NEWLINE_DELIMITED_JSON") {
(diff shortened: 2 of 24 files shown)


JulianUmbhau:
> Would you mind pointing me to the BigQuery docs for labels?
> 
> In the interest of using the simplest possible data structure, is there a reason to use a list instead of a named character vector?

Here are the docs: https://docs.cloud.google.com/bigquery/docs

And you are correct, the list was unnecessary. I have changed that to a named chr vector.

schifferl:
Adding my support for this feature – it is important for cost allocation in production applications that use R to communicate with BigQuery.

R/bq-perform.R
```diff
@@ -72,6 +72,8 @@ bq_perform_extract <- function(
   check_bool(print_header)
   check_string(billing)
 
+  labels <- check_labels(getOption("bigrquery.labels"))
```
hadley: As a general principle, I don't think it's good idea to allow global options to change what a computation does, only how it looks. I think it's probably ok in this case because I think you can frame this as a sort of logging operation, but it should appear in function arguments too.

I would also wonder if it would instead make sense to make it a field in `BigQueryConnection`. Would that also solve the problem for you or are you usually calling these low-level functions directly?

R/bq-perform.R
```diff
@@ -154,11 +159,16 @@ bq_perform_upload <- function(
...
+    configuration = list(
+      load = load,
+      labels = labels
+    )
+  )
   metadata <- bq_body(metadata, ...)
   metadata <- list(
     "type" = "application/json; charset=UTF-8",
-    "content" = jsonlite::toJSON(metadata, pretty = TRUE, digits = json_digits)
+    "content" = jsonlite::toJSON(metadata, auto_unbox = TRUE, pretty = TRUE)
```
hadley: Merge problem here?

R/utils.R
```diff
@@ -85,3 +85,29 @@ as_query <- function(x, error_arg = caller_arg(x), error_call = caller_env()) {
...
   is_installed("bigrquerystorage")
 }
+
+check_labels <- function(labels) {
+  if (is.null(labels) || length(labels) == 0 || (length(labels) == 1 && is.na(labels))) {
+    return(NULL)
+  }
+
+  if (!is.character(labels) || is.null(names(labels)) || anyNA(names(labels)) || any(names(labels) == "")) {
+    warning("Labels must be a named character vector; dropping labels", immediate. = TRUE, call. = FALSE)
```
hadley: In general, we prefer errors for data validation problems, not warnings. Can you please use `cli::cli_abort()` here instead?

R/utils.R
```diff
@@ -85,3 +85,29 @@ as_query <- function(x, error_arg = caller_arg(x), error_call = caller_env()) {
...
+    return(NULL)
+  }
+
+  nms <- names(labels)
+  bad_keys <- nms[nms != tolower(nms)]
+  if (length(bad_keys) > 0) {
```
hadley: I don't love this sort of very specific error message because it tightly couples bigrquery to the underlying API. If you remove them, do you still get a reasonably informative error message from the API call?

R/utils.R
```diff
@@ -85,3 +85,29 @@ as_query <- function(x, error_arg = caller_arg(x), error_call = caller_env()) {
 has_bigrquerystorage <- function() {
   is_installed("bigrquerystorage")
 }
+
+check_labels <- function(labels) {
+  if (is.null(labels) || length(labels) == 0 || (length(labels) == 1 && is.na(labels))) {
```
hadley: I don't think you should allow `NA` here; it's better to stick with just `NULL` if you want no labels.

R/utils.R
```diff
@@ -85,3 +85,29 @@ as_query <- function(x, error_arg = caller_arg(x), error_call = caller_env()) {
...
+    return(NULL)
+  }
+
+  if (!is.character(labels) || is.null(names(labels)) || anyNA(names(labels)) || any(names(labels) == "")) {
```
hadley: You could use `names2()` to simplify the logic here.

R/bq-parse.R
```diff
@@ -1,6 +1,6 @@
 bq_parse_single <- function(value, type, ...) {
   field <- bq_field("", type, ...)
-  field_j <- jsonlite::toJSON(as_json(field))
+  field_j <- jsonlite::toJSON(as_json(field), auto_unbox = TRUE)
```
hadley: Why was this change needed?

hadley (commented):
Thanks for working on this! A few comments on the overall approach and implementation below.

I assume you have tested this interactively?

R/utils.R
JulianUmbhau: You are correct, the error messages are information enough. I have removed the unnecessary checks.

R/utils.R
JulianUmbhau: Did not know that function, looks much better!

R/bq-perform.R
JulianUmbhau: Yes, fixed it.

R/bq-parse.R
JulianUmbhau: It is not, it was leftovers from testing around. Removing.

R/bq-perform.R
JulianUmbhau: I understand. 
I do use the low-level functions directly currently, but it is also a good point with a field in BigQueryConnection. 

Would it make sense to both:
1. Add labels field to BigQueryConnection
2. Add labels as arguments to each bq_perform_... function? - To set the labels explicitly

JulianUmbhau:
> Thanks for working on this! A few comments on the overall approach and implementation below.
> 
> I assume you have tested this interactively?

Ofc, thank you the comments!

We use this every day, but I have now tested it more thoroughly. 

I had to change the character vector back to list - i hadn't tested that change well enough.
The jsonlite::toJSON function drops the names of the character vector and changes it to an array. And the auto_unbox = TRUE is needed to ensure the value is not wrapped in an array.

R/bq-perform.R
hadley: Yes, I think so.

hadley: I'm also happy to do that myself, if you don't have the time.

hadley:
@JulianUmbhau could you please double check that my changes haven't broken anything for you?

hadley:
@JulianUmbhau I'm going to release early next week so I'd really appreciate a look if you have the time 😄

